PR #3, opened 2022-12-14
wip: refactor first draft

zeeshanlakhani:


.pre-commit-config.yaml
```diff
@@ -12,34 +12,34 @@ repos:
...
+        types: [rust]
+        pass_filenames: false
+      - id: clippy
+        name: clippy
+        description: Lint via clippy
+        entry: cargo clippy
+        language: system
+        args: ["--", "-D", "warnings"]
+        types: [rust]
+        pass_filenames: false
```
expede, line 29: haha yeah I guess I can just run `--no-verify`. It is nice to have the rest of them run while in WIP, though.

src/main.rs
```diff
@@ -1,245 +1,202 @@
...
+    io::{self, Cursor, Write},
+    str::{self, FromStr},
 };
 use url::Url;
 use wasmer::{imports, Function, Instance, Module, Store, Type, Value};
 
-/// Main entry point.
 #[tokio::main]
-async fn main() -> Result<(), Box<dyn Error>> {
+async fn main() -> Result<()> {
```
expede, line 34: Oh yeah, that's way easier 😅

src/main.rs
```diff
@@ -1,245 +1,202 @@
...
-    let file_content = match file_map {
-        Ok(x) => x.0,
-        Err(y) => panic!("no file {}", y),
-    };
+        Argument::Get { name } => {
+            let providers = client.get_providers(name.clone()).await?;
+            let providers = providers
+                .is_empty()
+                .then_some(providers)
+                .ok_or_else(|| anyhow!("Could not find provider for file {name}."))?;
```
expede, line 70: Why the reassignment? Can this not be...

```rust
let providers =
  client
    .get_providers(name.clone())
    .await?
    .is_empty()
    .then_some(providers)
    .ok_or_else(|| anyhow!("Could not find provider for file {name}."))?;
```

...or is that unidiomatic?

src/main.rs
```diff
@@ -1,245 +1,202 @@
...
-            })
-            .collect(),
-    );
-
-    let closure = workflow::closure::Closure {
-        resource,
-        action: workflow::closure::Action::from("wasm/run"),
-        inputs: workflow::closure::Input::IpldData(ipld_args),
-    };
+            println!("Fuck: {providers:?}");
```
expede, line 72: Just a `fixme`?

src/main.rs
```diff
@@ -1,245 +1,202 @@
...
-        match network_events.next().await {
-            Some(network::Event::InboundRequest { request, channel }) => {
-                if request == closure_cid2 {
-                    network_client.respond_file(res_copy.clone(), channel).await;
-                }
-            }
-            e => todo!("{:?}", e),
+            //
+        }
+        Argument::Provide { wasm, fun, args } => {
```
expede, line 89: This feels like a big block to just put inside a branch, no? Is there a cost to breaking this up and/or why not move it to another module?

src/main.rs
```diff
@@ -1,245 +1,202 @@
...
+                    .map(|a| match a {
+                        Value::I32(i) => Ipld::from(*i),
+                        _ => todo!(),
+                    })
+                    .collect(),
+            );
+
+            let closure = Closure {
+                resource,
+                action: Action::try_from("wasm/run")?,
```
expede, line 156: Why `try_from`? AFAICT the smart constructor should be unfailable, no?

src/workflow/receipt.rs
```diff
@@ -1,5 +1,5 @@
 use crate::db::schema::receipts;
-use diesel::prelude::*;
+use diesel::{Insertable, Queryable};
```
expede, line 2: 👍

src/workflow/task.rs
```diff
@@ -1,34 +1,42 @@
-//! A [Closure] wrapped with [Config]uration, metadata, and so on.
+//! A [Closure] wrapped with Configuration, metadata, and optional secret.
+
 use crate::workflow::{closure::Closure, config::Resources};
-use libipld::Ipld;
+use libipld::{serde as ipld_serde, Ipld};
```
expede, line 4: Just shorter than `libipld::serde`?

src/workflow/task.rs
```diff
@@ -1,34 +1,42 @@
...
 use std::collections::BTreeMap;
 
+const RESOURCES_KEY: &str = "resources";
+const META_KEY: &str = "meta";
+const SECRETS_KEY: &str = "secret";
+
+const WITH_KEY: &str = "with";
+const DO_KEY: &str = "do";
+const INPUTS_KEY: &str = "inputs";
```
expede, line 13: 👍 yeah good idea

expede: I feel like I'm converting between string types a lot 🤔 Is defaulting to `String` most places the wrong option?

src/workflow/task.rs
```diff
@@ -37,83 +45,46 @@ impl From<Closure> for Task {
...
-                    assoc.get("secret").ok_or(()).and_then(|ipld| match ipld {
-                        Ipld::Bool(b) => Ok(Some(*b)),
-                        Ipld::Null => Ok(None),
-                        _ => Err(()),
-                    })?;
+        let resources = if let Some(ipld) = map.get(RESOURCES_KEY) {
+            Resources::try_from(ipld)?
+        } else {
+            Resources::default()
+        };
```
expede, line 79: `let resources = if let Some(ipld) =`

Woah, that's gnarly... but idiomatic?

Could it be rephrased as something like...

```rust
let resources = map
    .get(RESOURCES_KEY)
    .map(Resources::try_from)
    .unwrap_or_else(Resources::default);
```

...or is that not commonly done?

.pre-commit-config.yaml
zeeshanlakhani: yep, but we'll get these working normally too (soon)

src/main.rs
zeeshanlakhani: there was a reason during cleanup, but I think you're right that it flows through :). I'll do this now.

src/main.rs
zeeshanlakhani: Z'isms,  yes!

src/main.rs
zeeshanlakhani: So, I'll put a todo here. I don't want the big block, but I was rethinking ways to abstract this out. So, I put it back in one spot before coming back to it.

src/main.rs
zeeshanlakhani: b/c this is fromstr under the hood, and the canonical approach seems to be tryfrom for this (and use fromstr at the bottom)

src/workflow/closure.rs
```diff
@@ -61,26 +60,28 @@ impl Into<Ipld> for Closure {
 }
 
 impl TryFrom<Ipld> for Closure {
-    type Error = ();
+    type Error = anyhow::Error;
```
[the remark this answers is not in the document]
zeeshanlakhani: I think it's good to leave this here b/c it cleans up results and error handling immensely elsewhere. We can then move to more specific errors later (via this error, or conversions to anyhow). Ideally, anyhow is the standard.

src/workflow/task.rs
zeeshanlakhani: So, pattern from wnfs :).

src/workflow/task.rs
zeeshanlakhani: it's a Rust thing. Going w/ String, means no consts and literals anyway sadly.

src/workflow/task.rs
```diff
@@ -37,83 +45,46 @@ impl From<Closure> for Task {
         Task {
             closure,
             metadata: Ipld::Null,
-            resources: Resources {
-                fuel: None,
-                time: None,
-            },
+            resources: Resources::default(),
```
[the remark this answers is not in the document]
zeeshanlakhani: def not a codesmell in Rust land and is common across most structs.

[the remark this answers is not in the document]
zeeshanlakhani: passing around None params is typically more "frowned upon".

src/workflow/task.rs
zeeshanlakhani: @expede yep, this is better (had this in the about to push)

src/workflow/closure.rs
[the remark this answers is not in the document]
expede: Is that so we're not converting between error types and wrapping them in aggregate errors and whatnot?

src/workflow/task.rs
zeeshanlakhani: even cleaner:

```rust
let resources = map
            .get(RESOURCES_KEY)
            .map_or_else(|| Ok(Resources::default()), Resources::try_from)?;
```

src/main.rs
expede: Huh, okay. Seems wasteful, but sure 🤷‍♀️

zeeshanlakhani: actually, yeah, the normal From is fine (i was pre-optimizing). will return.

src/main.rs
expede: I certainly have a lot of dangling stuff like this in my current code as well! Not a criticism or anything; just letting me inner 5 year old out. "Why? Why? Why? Why?" 😁

.pre-commit-config.yaml
expede: Just an FYI, my branches will probably have a bunch of `--no-verify` while they're WIP. I tend to push before e.g. a string of meetings, and a lot of those states won't compile.

I think Quinn has also disabled a bunch of these checks for the same reason.

expede: Would it make sense to push some of these to a GH Action?

zeeshanlakhani: I look at clippy as another extension of the type system :).

src/main.rs
zeeshanlakhani: @expede oh yeah, this won't work. That's due to how then_some expects the input.

src/workflow/closure.rs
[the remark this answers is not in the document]
zeeshanlakhani: yeah. you essentially want to use anyhow as your ground error type.

expede (approved):
Looks good to me! 🚀 Thanks!

PR #4, opened 2022-12-14
Wasm `NaN` Canonicalization & Metering

expede:
# Changelog

* Bump `cargo` deps
* Switch to Singlepass compiler
* Add metering
* Enable `NaN` canonicalization
* Per 📎Clippy, change all `Into`s into `From`s
* Use `anyhow` everywhere


src/main.rs
```diff
@@ -110,7 +114,22 @@ async fn main() -> Result<()> {
...
             .await?;
 
-            let mut store = Store::default();
+            let cost_function = |operator: &Operator| -> u64 {
+                match operator {
+                    Operator::LocalGet { .. } | Operator::I32Const { .. } => 1,
+                    Operator::I32Add { .. } => 2,
+                    _ => 0,
+                }
+            };
```
expede, line 123: The gas schedule goes here. This is just an example copy/pasta'd from the Wasmer middleware repo

src/main.rs
```diff
@@ -110,7 +114,22 @@ async fn main() -> Result<()> {
...
+
+            let metering_middleware = Arc::new(Metering::new(10, cost_function));
+
+            let mut basic_compiler = Cranelift::new();
+            let compiler_config = basic_compiler.canonicalize_nans(true);
```
expede, line 128: Screenshot 2022-12-14 at 14 39 37

src/wasm/operator.rs
```diff
@@ -0,0 +1,9 @@
+use wasmer::wasmparser::Operator;
+
+pub fn to_cost(op: &Operator) -> u64 {
+    match op {
+        Operator::LocalGet { .. } | Operator::I32Const { .. } => 1,
+        Operator::I32Add { .. } => 2,
+        _ => 0,
```
expede, line 7: This will not be `0`. We may just pilfer the FVM schedule to get started 😉

src/workflow/closure.rs
```diff
@@ -37,24 +37,25 @@ pub struct Closure {
...
-    fn into(self) -> Link<Closure> {
-        let mut closure_bytes = Vec::new();
-        <Closure as Into<Ipld>>::into(self).encode(DagCborCodec, &mut closure_bytes);
+impl TryFrom<Closure> for Link<Closure> {
+    type Error = anyhow::Error;
 
-        Link::new(Cid::new_v1(
+    fn try_from(closure: Closure) -> Result<Link<Closure>, Self::Error> {
+        let mut closure_bytes = Vec::new();
+        <Closure as Into<Ipld>>::into(closure).encode(DagCborCodec, &mut closure_bytes)?;
```
expede, line 45: I hate that the compiler needs me to write out `<Closure as Into<Ipld>>::into(closure)` with the burning passion of a thousand suns. Is there a better way?

src/workflow/closure.rs
```diff
@@ -37,24 +37,25 @@ pub struct Closure {
...
-    fn into(self) -> Ipld {
+impl From<Closure> for Ipld {
+    fn from(closure: Closure) -> Self {
         Ipld::Map(BTreeMap::from([
-            ("with".to_string(), Ipld::String(self.resource.into())),
-            ("do".to_string(), self.action.into()),
-            ("inputs".to_string(), self.inputs.into()),
+            ("with".to_string(), Ipld::String(closure.resource.into())),
+            ("do ".to_string(), closure.action.into()),
+            ("inputs".to_string(), closure.inputs.into()),
```
expede, line 58: Below there are many complaints from the checker. I think I disagree with all of them? The code seems reasonable, no? I guess it is relatively dense, but that's not what it's complaining about. Would it prefer these broken up across multiple variables or something?

expede: Well, they're gone. I have no idea. I didn't change anything 🤷‍♀️

src/workflow/invocation.rs
```diff
@@ -31,85 +34,82 @@ impl Into<Ipld> for Invocation {
...
+                            _ => bail!("Not a link"),
                         })
                     }
-                    _ => Err(()),
-                }?;
-
-                Ok(Invocation { meta, prf, run })
-            }
-            _ => Err(()),
+                    other => bail!("Expected a List, but got something else: {:?}", other),
```
expede, line 58: I threw a bunch of `{:?}` in because the conversions to normal strings was... not as simple as I wanted. This is probably overkill 🤷‍♀️

PR #6, opened 2022-12-16
refactor: work through conversions

zeeshanlakhani:


src/workflow/closure.rs
```diff
@@ -41,21 +47,21 @@ impl TryFrom<Closure> for Link<Closure> {
     type Error = anyhow::Error;
 
     fn try_from(closure: Closure) -> Result<Link<Closure>, Self::Error> {
-        let mut closure_bytes = Vec::new();
-        <Closure as Into<Ipld>>::into(closure).encode(DagCborCodec, &mut closure_bytes)?;
```
zeeshanlakhani, line 45: @expede was able to remove this.

expede: Oh yeah, that makes sense!

src/network/client.rs
```diff
@@ -20,13 +19,10 @@ impl Client {
     ) -> Result<(Self, mpsc::Receiver<Event>, EventLoop)> {
         let (command_sender, command_receiver) = mpsc::channel(1);
         let (event_sender, event_receiver) = mpsc::channel(1);
-        let keypair = Keypair::generate_ed25519();
```
zeeshanlakhani, line 23: will bring back the ability to start with a seed in another pr.

src/workflow/invocation.rs
```diff
@@ -90,26 +96,15 @@ impl TryFrom<Ipld> for Batch {
...
-            _ => bail!("Can only convert from a map"),
-        }
+        let map = from_ipld::<BTreeMap<String, Ipld>>(ipld)?;
+        let flow = map
+            .iter()
+            .try_fold(BTreeMap::new(), |mut acc, (key, value)| {
+                let task = Task::try_from(value)?;
+                acc.insert(TaskLabel::new(key.to_string()), task);
+                Ok::<_, anyhow::Error>(acc)
+            })?;
```
expede, line 106: Oh yeah, that's definitely cleaner ✨ I'm still getting used to what the right patterns for this stuff is. Glad to see that folds are idiomatic!

zeeshanlakhani: yep. the folds/try_folds over iters are good, just need that silly type annotation!

Cargo.toml
```diff
@@ -17,7 +17,7 @@ authors = ["Brooklyn Zelenka <hello@brooklynzelenka.com>"]
 [lib]
 path = "src/lib.rs"
 bench = false
-doctest = false
+doctest = true
```
zeeshanlakhani, line 20: @expede now doctests will run as part of norm tests.

src/workflow/pointer.rs
```diff
@@ -13,15 +17,15 @@ pub struct Promise {
...
-            Some(Status::Success) => "ucan/ok".to_string(),
-            Some(Status::Failure) => "ucan/err".to_string(),
-            None => "ucan/promise".to_string(),
+        let key: String = match promise.result {
+            Some(Status::Success) => OK_BRANCH.to_string(),
+            Some(Status::Failure) => ERR_BRANCH.to_string(),
+            None => PTR_BRANCH.to_string(),
         };
 
         Ipld::Map(BTreeMap::from([(key, promise.invoked_task.into())]))
```
expede, line 31: Minor suggestion:

```suggestion
impl From<Promise> for Ipld {
    fn from(promise: Promise) -> Self {
        let key: &str = match promise.result {
            Some(Status::Success) => OK_BRANCH,
            Some(Status::Failure) => ERR_BRANCH,
            None => PTR_BRANCH,
        };

        Ipld::Map(BTreeMap::from([(key.to_string(), promise.invoked_task.into())]))
```

src/workflow/pointer.rs
```diff
@@ -13,15 +17,15 @@ pub struct Promise {
...
-    pub branch_selector: Option<Status>,
+    pub result: Option<Status>,
 }
 
 impl From<Promise> for Ipld {
     fn from(promise: Promise) -> Self {
-        let key: String = match promise.branch_selector {
-            Some(Status::Success) => "ucan/ok".to_string(),
-            Some(Status::Failure) => "ucan/err".to_string(),
-            None => "ucan/promise".to_string(),
```
zeeshanlakhani, line 24: spec referenced this as ucan/ptr.

src/workflow/pointer.rs
zeeshanlakhani: gonna make one more minor change too.

src/workflow/pointer.rs
```diff
@@ -111,44 +117,54 @@ impl TryFrom<Ipld> for InvokedTaskPointer {
...
+                } else {
+                    Err(anyhow!("Unexpected format for local InvokedTaskPointer"))
+                }
+            }
+            [Ipld::Link(ptr), Ipld::String(label)] => Ok(InvokedTaskPointer {
+                invocation: InvocationPointer::Remote(*ptr),
+                label: TaskLabel(label.to_string()),
+            }),
+
+            _ => Err(anyhow!("Unexpected number of segments in IPLD tuple")),
```
expede, line 141: Ah, yeah, avoiding the hidden `return` makes sense 👍

Cargo.toml
expede: Oooh gotcha. Nice 👍

Cargo.toml
```diff
@@ -58,7 +57,7 @@ ucan = "0.1"
 url = "2.3"
 wasmer = { version = "3.1", features = ["compiler"] }
 wasmer-compiler-singlepass = "3.1.0"
-wasmer-middlewares = "3.1.0"
```
expede, line 61: Is there a way to prevent `cargo add` from doing this? Or is the problem that I'm using `cargo add`? 😅

src/network/client.rs
expede: Sure, yeah. I'm not sure that it's a critical feature 🤷‍♀️

src/workflow/invocation.rs
expede: Is the `ControlFlow` stuff not worth it I guess? Probably fast enough just to return a bunch of `Err` in quick succession if you end up on that track?

zeeshanlakhani: I didn't see the need here as we can bubble up. I'd use it if there's a result I really want to break with and use, cause then you have to match on the ControlFlow ops at the call site.

expede (approved):
Looks great! Thanks for demonstrating all of those stylistic fixes 😀 🚀

PR #30, opened 2023-01-18
Adds gossippub/mdns to swarm behaviour for peering and topic subscriptions/publishings

zeeshanlakhani:
# Description

Includes:

* updates around db receipts
* removal of .env since we have direnv/.envrc
* moves to libp2p master (which has a ton of changes in the pipe)
* tokio for all runtime/swarming
* updates outdated libs


zeeshanlakhani:
will open a pr under a diff name branch shortly.

src/db.rs
```diff
@@ -1,12 +1,9 @@
 pub mod schema;
 
 use diesel::prelude::*;
-use dotenvy::dotenv;
 use std::env;
 
 pub fn establish_connection() -> SqliteConnection {
-    dotenv().ok();
```
expede, line 8: Are you assuming that the developer has direnv enabled?

zeeshanlakhani: I do if you're developing with flake, no?

PR #31, opened 2023-01-19
Adds gossippub/mdns to swarm behaviour for peering and topic subscriptions/publishings

### LINKED ISSUE #27: impl gossipsub for receipt topic pubsub layer
# Summary

Use gossipsub (and mdns) to share receipts (possibly other data) with connected peers.

zeeshanlakhani:
# Description

Includes:

* updates around db receipts
* moves to libp2p master (which has a ton of changes in the pipe)
* tokio for all runtime/swarming
* direct floodsub capability too (if not trying to force mesh mechanics)
* updates outdated libs

Closes #27 for now.


src/main.rs
```diff
@@ -42,7 +40,12 @@ async fn main() -> Result<()> {
 
     let opts = Args::parse();
     let keypair = Keypair::generate_ed25519();
-    let swarm = swarm::build_swarm(keypair).await?;
+
+    let mut swarm = swarm::new(keypair).await?;
+
+    // subscribe to `receipts` topic
+    swarm.behaviour_mut().gossip_subscribe(RECEIPTS_TOPIC)?;
```
expede, line 47: Well, that was easy 🙌

src/main.rs
```diff
@@ -71,10 +74,10 @@ async fn main() -> Result<()> {
...
         Argument::Get { name } => {
             let providers = client.get_providers(name.clone()).await?;
-            let providers = providers
-                .is_empty()
-                .then_some(providers)
-                .ok_or_else(|| anyhow!("Could not find provider for file {name}."))?;
+
+            if providers.is_empty() {
+                Err(anyhow!("Could not find provider for file {name}."))?;
+            }
```
expede, line 80: 👍

src/main.rs
```diff
@@ -164,33 +166,46 @@ async fn main() -> Result<()> {
...
             let mut closure_bytes = Vec::new();
             closure_ipld
                 .encode(DagCborCodec, &mut closure_bytes)
                 .expect("CBOR Serialization");
 
-            let closure_cid = Cid::try_from(closure_bytes)?
+            let link: Link<Closure> = Closure::try_into(closure)?;
+            let closure_cid = link
                 .to_string_of_base(Base::Base32HexLower)
                 .expect("string CID");
```
expede, line 179: I have no problem with this being reworked, but could you explain a bit about what the advantage of allocating another binding here is? I don't see `link` used anywhere else 🤔

src/network/eventloop.rs
```diff
@@ -1,21 +1,39 @@
...
 use tokio::sync::{mpsc, oneshot};
+use void::Void;
+
+type HandlerErr = Either<
+    Either<
+        Either<Either<GossipsubHandlerError, ConnectionHandlerUpgrErr<CodecError>>, io::Error>,
+        Void,
+    >,
+    ConnectionHandlerUpgrErr<io::Error>,
+>;
```
expede, line 34: Why the deeply nested `Either` instead of a custom error type and a way to lift errors into the sum?

src/network/eventloop.rs
```diff
@@ -121,7 +190,7 @@ impl EventLoop {
                 }
             },
             SwarmEvent::Behaviour(ComposedEvent::RequestResponse(
-                RequestResponseEvent::OutboundFailure {
```
expede, line 124: Are these just type aliases for legibility when used without the module?

src/network/eventloop.rs
```diff
@@ -44,27 +62,49 @@ impl EventLoop {
...
-    async fn handle_event(
-        &mut self,
-        event: SwarmEvent<
-            ComposedEvent,
-            EitherError<ConnectionHandlerUpgrErr<io::Error>, io::Error>,
-        >,
-    ) {
+    async fn handle_event(&mut self, event: SwarmEvent<ComposedEvent, HandlerErr>) {
         match event {
+            SwarmEvent::Behaviour(ComposedEvent::Floodsub(FloodsubEvent::Message(message))) => {
```
expede, line 80: Why support FloodSub?

src/main.rs
```diff
@@ -164,33 +166,46 @@ async fn main() -> Result<()> {
...
 
             let mut conn = db::establish_connection();
 
-            let new_receipt = NewReceipt {
+            let new_receipt = Receipt {
+                id: Uuid::new_v4().to_string(),
```
expede, line 184: Why not let the DB just give you a sequential ID?

expede: I guess our other option down the road is to use the CID once there's a signature and nonce, so yeah, okay sure this works for now 👍

tests/integration_test.rs
```diff
@@ -1 +1,4 @@
-
+#[test]
+fn test_add() {
+    assert!(true);
+}
```
expede, line 4: Still WIP?

expede:
@zeeshanlakhani are you able to successfully run `nix develop` with these changes? If so, can you try leaving the directory, entering it again, and running `nix develop` (in case `direnv` has cached something)? I'm getting the following hash verification failure:

```
Latte:~/D/S/i/ipvm on  zl/gossippub+ [$] is 📦 v0.1.0 via 🦀 took 9s
»  nix develop
error: hash mismatch in fixed-output derivation '/nix/store/vsrjhpadr6j7z9rkixadxvbisdmmq8gc-source.drv':
         specified: sha256-AMOND5q1XzNhN5smVJp+2sGl/OqbxkGPGuPBCE48Hik=
            got:    sha256-tqNTh8ojTT80kl9jDCMq289jSfPz82gQ37UFAmnfsOw=
error: 1 dependencies of derivation '/nix/store/zj5c28qswng9ryppdi3d6dblc2dlxfn6-irust-1.66.0.drv' failed to build
error: 1 dependencies of derivation '/nix/store/klcj710bl4qgz0az9h720c6cjd385d20-ipvm-env.drv' failed to build
```

flake.nix
```diff
@@ -70,7 +71,7 @@
 
       packages.irust = pkgs.rustPlatform.buildRustPackage rec {
         pname = "irust";
-        version = "1.65.1";
+        version = "1.66.0";
```
expede, line 74: I think that the sha256 needs to get bumped as well

```
✗  nix develop
error: hash mismatch in fixed-output derivation '/nix/store/vsrjhpadr6j7z9rkixadxvbisdmmq8gc-source.drv':
         specified: sha256-AMOND5q1XzNhN5smVJp+2sGl/OqbxkGPGuPBCE48Hik=
            got:    sha256-tqNTh8ojTT80kl9jDCMq289jSfPz82gQ37UFAmnfsOw=
error: 1 dependencies of derivation '/nix/store/zj5c28qswng9ryppdi3d6dblc2dlxfn6-irust-1.66.0.drv' failed to build
error: 1 dependencies of derivation '/nix/store/klcj710bl4qgz0az9h720c6cjd385d20-ipvm-env.drv' failed to build
```

src/main.rs
zeeshanlakhani: Yep, I plan on the latter actually, but also to support sequential ID, we won't have varchar. But yeah, plan to use the cid once there's a sig + nonce.

tests/integration_test.rs
zeeshanlakhani: Yep. No integration tests, yet, but the file can't be empty.

src/network/eventloop.rs
zeeshanlakhani: There are some benefits if we want a direct pubsub vs a mesh w/ gossipsub. This is mainly for testing ground purposes right now, as there were some interesting mesh issues I ran into before getting things to work concretely. Once we're ready for a real "version" of things, we can probably cut it, but it's good for experimentation.

src/network/eventloop.rs
zeeshanlakhani: libp2p is removing them essentially.

src/main.rs
zeeshanlakhani: I broke this from your original code, as I was running up the "demo" in its entirety now. Cid (from the multiformat library) will throw an error on `try_into` (when given `closure_bytes`) as it can't decipher the version. This is to get the cid that we need.

src/network/eventloop.rs
zeeshanlakhani: This is due to the underlying swarmevent errors (and what I was using from examples). The easier approach is a generic, which, I'll push.

src/network/eventloop.rs
zeeshanlakhani: so these alias's no longer exist, unless we wanted to create ourselves.

src/main.rs
zeeshanlakhani: But, something I'd like to re-explore in how it's derived.

zeeshanlakhani: This is leveraging the free "into" that comes from TryFrom here in closure.rs:

```rust
impl TryFrom<Closure> for Link<Closure> {
    type Error = anyhow::Error;

    fn try_from(closure: Closure) -> Result<Link<Closure>, Self::Error> {
        let ipld: Ipld = closure.into();
        let bytes = DagCborCodec.encode(&ipld)?;
        Ok(Link::new(Cid::new_v1(
            DagCborCodec.into(),
            Code::Sha3_256.digest(&bytes),
        )))
    }
}
```

PR #55, opened 2023-02-18
refactor: move wasm to separate crate, cleanup owns/clones/refs

zeeshanlakhani:
refactor: move wasm to separate crate, cleanup owns/clones/refs/execution

Includes:

* Move execution to specialized wasmtime env struct, allowing for
expansion on linker, store.
* Instantiation and execution is now as simple as:

```rust
let mut env = wasmtime::World::instantiate(wasm_bytes, fun, state).await?;
let res = env.execute(ipld_args.clone()).await?;
```


zeeshanlakhani:
For posterity: https://github.com/ipvm-wg/ipvm/actions/runs/4209589197/jobs/7306632704#step:8:871 will be fixed in the new nightly out next week (it's in the mainline branch).

Issue #163, opened 2023-06-22
Add option to import keypair to runtime at startup for libp2p network/swarm setup


zeeshanlakhani:
### Summary

Give the option to import a keypair at startup for libp2p network/swarm setup and/or to use a secret seed even to generate one internally. Specified types of [digital signature](https://en.wikipedia.org/wiki/Digital_signature) scheme are here: https://github.com/libp2p/rust-libp2p/blob/master/identity/src/keypair.rs#L102.

### Goals

Add configuration to be able to **either** provide:

- [ ] A secret key seed akin to https://github.com/libp2p/rust-libp2p/tree/master/examples/file-sharing#usage for swarm/transport setup; or,
- [ ] A means to load a key file path knowing that the node has the private key of the keypair, using `Keypair::from`; or, 
- [ ] Default to what we have today, which generates a keypair at startup.

zeeshanlakhani:
Closed by #172

PR #172, opened 2023-06-28
Add pubkey importing / seed generation

mriise:
# Description
Add support for importing ed25519 keypairs and generating a new keypair from seed bytes.

## Link to issue

#163 

## Type of change

- [x] This change requires a documentation update

## Test plan (required)

TODO

Demonstrate the code is solid. Which commands did you test with and what are the expected results?
Which tests have you added or updated? Do the tests cover all of the changes included in this PR?

## Screenshots/Screencaps

Please add previews of any UI Changes.


homestar-runtime/src/settings.rs
```diff
@@ -22,6 +24,44 @@ pub struct Monitoring {
...
 }
 
+#[derive(Default, Clone, Debug, Deserialize)]
+/// Configure how the Network keypair is generated or using an existing one
+pub(crate) enum PubkeyConfig {
+    #[default]
+    Random,
+    /// Seed bytes are hashed with SHA-256 to produce the ed25519 secret key.
+    RandomWithSeed(Vec<u8>),
+    Existing(Vec<u8>),
```
zeeshanlakhani, line 34: Though internally we can treat this as `Vec<u8>` @mriise, as far as configuration goes, should this come from a file path or something similar? Typically, the config will be pass in as toml/yaml, etc, and bytes won't be listed directly.

zeeshanlakhani (commented):
One comment thus far, but awesome to see this @mriise! We should also have this PR use the setting upon swarm startup, right?

zeeshanlakhani:
@mriise if you install the pre-commit hook and such, you'll get the cargo fmt stuff for free too. LMK if you want to chat through that or the nix flake.

homestar-runtime/src/settings.rs
mriise: Yep! this is changed in the newest commit!

homestar-runtime/Cargo.toml
```diff
@@ -59,8 +59,10 @@ openssl = { version = "0.10", features = ["vendored"] }
 proptest = { version = "1.2", optional = true }
 reqwest = { version = "0.11", features = ["json"] }
 semver = "1.0"
+
```
zeeshanlakhani, line 62: minor: unnecessary spaces.

homestar-runtime/Cargo.toml
```diff
@@ -69,6 +71,8 @@ tracing-logfmt = "0.3"
 tracing-subscriber = { version = "0.3", features = ["env-filter", "parking_lot", "registry"] }
 tryhard = "0.5"
 url = "2.4"
+sha2 = "0.10.7"
+pem = "2.0.1"
```
zeeshanlakhani, line 75: same as above.

homestar-runtime/config/testkey.pem
```diff
@@ -0,0 +1,3 @@
+-----BEGIN PRIVATE KEY-----
```
zeeshanlakhani, line 1: Let's put this in a fixtures repo, akin to homestar-wasm.

homestar-runtime/src/settings.rs
```diff
@@ -22,6 +24,55 @@ pub struct Monitoring {
     process_collector_interval: u64,
 }
 
+#[derive(Default, Clone, Debug, Deserialize)]
+/// Configure how the Network keypair is generated or using an existing one
+pub(crate) enum PubkeyConfig {
+    #[default]
+    #[serde(rename = "random")]
+    Random,
```
zeeshanlakhani, line 32: To your point in chat @mriise, I'd be ok removing `generatefromseed` for just using a seed for all homestar nodes internally that is our default. What do you think?  Other systems, mainly blockchain-related stuff using libp2p, always gave the option, but I'm not stuck to it.

homestar-runtime/src/settings.rs
```diff
@@ -22,6 +24,55 @@ pub struct Monitoring {
...
+    GenerateFromSeed(String),
+    /// File path to a PEM encoded ed25519 key
+    #[serde(rename = "path")]
+    Existing(String),
+}
+
+
+impl PubkeyConfig {
+    /// Produce a Keypair using the given configuration. Consumes `self` to avoid keeping secrets laying around.
+    fn generate_keypair(self) -> Result<identity::Keypair, DecodingError> {
```
zeeshanlakhani, line 44: instead of decoding error, let's use `anyhow::Result` as per the runtime. I want to start labeling errors and bubbling up to anyhow, but we can tackle that in an error-specific PR. If you use anyhow here, you can just use `?` or `anyhow::context` to provide the decodingerror as a contextual error chain.

homestar-runtime/src/settings.rs
```diff
@@ -22,6 +24,55 @@ pub struct Monitoring {
...
+        match self {
+            PubkeyConfig::Random => Ok(identity::Keypair::generate_ed25519()),
+            PubkeyConfig::GenerateFromSeed(seed) => {
+                let mut hasher = Sha256::default();
+                hasher.update(&seed);
+                identity::Keypair::ed25519_from_bytes(hasher.finalize())
+            },
+            PubkeyConfig::Existing(path) => {
+                let path = Path::new(&path);
+                println!("{:?}", path);
```
zeeshanlakhani, line 54: guessing prints are just for testing. We do have `tracing` enabling for logging.

homestar-runtime/src/settings.rs
```diff
@@ -22,6 +24,55 @@ pub struct Monitoring {
...
+                let pem_file = 
+                {  
+                    let mut s = String::new();
+                    // TODO convert err
+                    std::fs::File::open(path).unwrap().read_to_string(&mut s).unwrap();
```
zeeshanlakhani, line 59: see above w/ removing unwraps for `?` with anyhow.

homestar-runtime/src/settings.rs
```diff
@@ -22,6 +24,55 @@ pub struct Monitoring {
...
+                    s
+                };
+                // TODO convert err
+                let pem = pem::parse(pem_file).unwrap();
+                // we only take ed25519
+                if pem.tag() != "PRIVATE KEY" {
+                    // TODO custom error at this point... 
+                    panic!("Not a private key pem file")
```
zeeshanlakhani, line 67: We don't want to panic here, but let all this bubble up and use `expect` upon loading settings in `main.rs`.

homestar-runtime/src/settings.rs
```diff
@@ -133,3 +187,16 @@ impl Settings {
...
+
+#[cfg(test)]
+mod test {
+    use crate::Settings;
+
+    #[test]
+    fn load_settings() {
+        let settings = Settings::load().unwrap();
+
+        println!("{:?}", settings.node.network.keypair_config.generate_keypair())
```
zeeshanlakhani, line 199: minor: guessing we'll have a real assert here, so just mentioning it aloud.

homestar-runtime/src/settings.rs
```diff
@@ -22,6 +24,55 @@ pub struct Monitoring {
...
+                let pem_file = 
+                {  
+                    let mut s = String::new();
+                    // TODO convert err
+                    std::fs::File::open(path).unwrap().read_to_string(&mut s).unwrap();
+                    s
+                };
+                // TODO convert err
+                let pem = pem::parse(pem_file).unwrap();
+                // we only take ed25519
```
zeeshanlakhani, line 64: Must we only take `ed25519` btw? We could support others, i.e. https://libp2p.github.io/rust-libp2p/libp2p_identity/index.html? Maybe this is part of the configuration.

homestar-runtime/config/settings.toml
```diff
@@ -2,3 +2,19 @@
 process_collector_interval = 10
 
 [node]
+
+[node.network]
```
zeeshanlakhani, line 6: I say we don't need this, as defaults are merged in anyway. For testing, we can just load settings and set the keypair_config or load a fixtures config with just the change (rest of defaults will make their way through anyway, as per config.rs).

homestar-runtime/src/settings.rs
mriise: fair enough to have it though, its a bit silly- but since we give the option of random, specifying a different seed is important enough.

homestar-runtime/src/settings.rs
mriise: yep, will add logs for tracing about the same time as the actual tests

homestar-runtime/src/settings.rs
mriise: actual tests to be a thing are still in the works

homestar-runtime/src/settings.rs
mriise: we can support Secp256k1 keys as it will likely be need for blockchains, but I would prefer to stay away from RSA. Am neutral on the random variant.

homestar-runtime/config/settings.toml
mriise: There is an issue with the defaults merging atm since `load_settings` will fail without _all_ fields specified

homestar-runtime/src/settings.rs
zeeshanlakhani: Yep. Good with that/those choice(es)

homestar-runtime/config/settings.toml
zeeshanlakhani: I'll double-check on this and fix if so.

zeeshanlakhani: Think it was missing the default, that you added :).

mriise: issue was with serde not filling in the defaults.

homestar-runtime/src/settings.rs
mriise: added support for secp256k1, however importing is a bit weird. for the moment i am just importing a raw key bytes in PEM format, but it should probably be pkcs#8 or something.

homestar-runtime/config/settings.toml
zeeshanlakhani: > issue was with serde not filling in the defaults.

Yep. Just missed this.

zeeshanlakhani: Of note @mriise, my next PR will clean this up for settings across the board, now with tests :).

homestar-runtime/Cargo.toml
```diff
@@ -53,14 +53,17 @@ ipfs-api = { version = "0.17", optional = true }
 ipfs-api-backend-hyper = { version = "0.6", features = ["with-builder", "with-send-sync"], optional = true }
 itertools = "0.10"
 libipld = { workspace = true }
-libp2p = { version = "0.51", default-features = false, features = ["kad", "request-response", "macros", "identify", "mdns", "gossipsub", "tokio", "dns", "mplex", "tcp", "noise", "yamux", "websocket"] }
+libp2p = { version = "0.51", default-features = false, features = ["kad", "request-response", "macros", "identify", "mdns", "gossipsub", "tokio", "dns", "mplex", "tcp", "noise", "yamux", "websocket", "ed25519", "secp256k1"] }
 libsqlite3-sys = { version = "0.26", features = ["bundled"] }
 openssl = { version = "0.10", features = ["vendored"] }
 proptest = { version = "1.2", optional = true }
+rand = "0.8.5"
```
zeeshanlakhani, line 60: minor: we'll let patch versions ride. so:

```suggestion
rand = "0.8"
```

homestar-runtime/Cargo.toml
```diff
@@ -53,14 +53,17 @@ ipfs-api = { version = "0.17", optional = true }
...
 reqwest = { version = "0.11", features = ["json"] }
+sec1 = { version = "0.7.2", features = ["pem", "der"] }
```
zeeshanlakhani, line 62: same as above, w.r.t. version

homestar-runtime/Cargo.toml
```diff
@@ -53,14 +53,17 @@ ipfs-api = { version = "0.17", optional = true }
...
 semver = "1.0"
 serde = { version = "1.0", features = ["derive"] }
 serde_ipld_dagcbor = { workspace = true }
+serde_with = { version = "*", features = ["base64"] }
```
zeeshanlakhani, line 66: minor: I brought in serde_with in my PR too, we should version this vs `*`.

homestar-runtime/fixtures/settings-import-secp256k1.toml
```diff
@@ -0,0 +1,8 @@
+[monitoring]
+process_collector_interval = 10
+
+[node]
+
+[node.network.keypair_config]
+# `openssl ecparam -genkey -name secp256k1 -outform DER -o secp256k1_key.der`
```
zeeshanlakhani, line 7: so great to have the docs/comments here!

homestar-runtime/src/network/swarm.rs
```diff
@@ -19,7 +18,11 @@ use std::{fmt, time::Duration};
 
 /// Build a new [Swarm] with a given transport and a tokio executor.
 pub async fn new(settings: &settings::Node) -> Result<Swarm<ComposedBehaviour>> {
-    let keypair = Keypair::generate_ed25519();
+    let keypair = settings
+        .network
+        .keypair_config
+        .generate_keypair()
+        .with_context(|| "Failed to generate/import keypair for libp2p".to_string())?;
```
zeeshanlakhani, line 25: minor: have been not capitalizing error strings (as per how rust typically does it too).

homestar-runtime/src/settings.rs
```diff
@@ -22,6 +30,115 @@ pub struct Monitoring {
...
+            PubkeyConfig::Random => Ok(identity::Keypair::generate_ed25519()),
+            PubkeyConfig::GenerateFromSeed(PupkeyRNGSeed { key_type, seed }) => {
+                // seed RNG with supplied seed
+                let mut r = rand::prelude::StdRng::from_seed(*seed);
+                let mut new_key: [u8; 32] = r.gen();
+
+                match key_type {
+                    KeyType::Ed25519 => {
+                        identity::Keypair::ed25519_from_bytes(new_key).map_err(|e| {
+                            anyhow!("Failed to generate ed25519 key from random: {:?}", e)
```
zeeshanlakhani, line 87: minor: `{:#?}` instead of `{:?}` will pretty-print debug logs.

zeeshanlakhani (commented):
Some super minor comments @mriise, but this is shaping up really nice.

homestar-runtime/Cargo.toml
mriise: this being left in was a mistake sorry

homestar-runtime/src/network/swarm.rs
mriise: sounds good

homestar-runtime/src/settings.rs
mriise: yeah we can pretty print these. though if we are going to be showing these errors to a lib that displays errors better in general like https://docs.rs/color-eyre/latest/color_eyre/

zeeshanlakhani: @mriise yep, we'll prob bring in color_eyre for local dev. Nonetheless, this is more so for logs that will be picked up by grafana/other-simialr tool (or logging service), especially as we use logfmt.

homestar-runtime/src/settings.rs
```diff
@@ -133,3 +266,42 @@ impl Settings {
...
+            .unwrap()
+            .public()
+            .verify(msg, &signature));
+    }
+
+    #[test]
+    fn import_secp256k1_key() {
+        let settings = Settings::build("fixtures/settings-import-secp256k1.toml".into()).unwrap();
+
+        settings.node.network.keypair_config.keypair().unwrap();
```
zeeshanlakhani, line 298: minor: 

```suggestion
assert!(settings.node.network.keypair_config.keypair().is_ok())
```

homestar-runtime/src/settings.rs
```diff
@@ -133,3 +266,42 @@ impl Settings {
...
+    }
+
+    #[test]
+    fn seeded_secp256k1_key() {
+        let settings = Settings::build("fixtures/settings-random-secp256k1.toml".into()).unwrap();
+
+        settings.node.network.keypair_config.keypair().unwrap();
```
zeeshanlakhani, line 305: same as above.

homestar-runtime/src/settings.rs
```diff
@@ -22,6 +31,126 @@ pub struct Monitoring {
...
+                let path = Path::new(&path);
+                let mut file = std::fs::File::open(path).context("unable to read key file")?;
+
+                let mut buf = Vec::new();
+                file.read_to_end(&mut buf)
+                    .context("unable to read bytes from file, is the file corrupted?")?;
+
+                match key_type {
+                    KeyType::Ed25519 => {
+                        const PEM_HEADER: &str = "PRIVATE KEY";
```
zeeshanlakhani, line 119: minor: I tend to put consts up top, though doesn't matter much.

zeeshanlakhani (approved):
Great work @mriise. I had a super minor comment on the tests below, i.e. assert vs unwrap. 

With that, this is mergeable! => First PR in the books!!!!

homestar-runtime/src/settings.rs
mriise: Didn't want it conflated when we have the secp PEM headers, so would pref to keep as is.

homestar-runtime/src/settings.rs
mriise: panic on unwrap arguably give better info on failure than an assert of `is_ok` here since the unwrap will print out the actual error and `is_ok` will just print the bool

zeeshanlakhani: @mriise thinking about it, then `expect` is what we'd want (as it will also print the actual error, and the context for the error given).

zeeshanlakhani:
Great work @mriise. Last thing, @mriise can you sign your commits? May be easier to rebase here and sign the singular commit. Thanks!

PR #180, opened 2023-07-04
refactor: first-pass at runtime interface, signals, shutdown, and more

zeeshanlakhani:
This work precedes subsequent work on running workflows from the Homestar runner/server/runtime but sets everything into gear.

Includes:

* overall refactoring of events and event handling, streamlining how to work with senders, etc
* introduce a runner which can
  - interact with task/batch specific-abort handles introduced in joinsets within a worker
  - interact with commands, which will happen through the CLI interface, i.e. star, run-workflow, akin to something like temporal-server
  - start-up event handler, websocket server, takes in db and can mock via implementation
* lots of cleanups


zeeshanlakhani:
@mriise will hold off merging this until #172 is in.

PR #201, opened 2023-07-17
Swarm Bootstrap Nodes

mriise:
# Description

Adds bootstrap nodes in settings that will be connected to on node startup.

## Link to issue

Please add a link to any relevant issues/tickets.

## Type of change

Please delete options that are not relevant.

## Test plan (required)

Parsing testing is added with the other settings test. Tests actually dialing nodes will be added in a different PR since this is a perquisite to gossip/receipt testing anyways.

### DIFF
--- a/homestar-runtime/fixtures/settings.toml
+++ b/homestar-runtime/fixtures/settings.toml
@@ -6,3 +6,4 @@ process_collector_interval = 10
 [node.network]
 events_buffer_len = 1000
 websocket_port = 9999
+bootstrap_addresses = ["/ip4/127.0.0.1/tcp/9998/ws/"]
--- a/homestar-runtime/fixtures/settings.toml
+++ b/homestar-runtime/fixtures/settings.toml
@@ -6,4 +6,4 @@ process_collector_interval = 10
 [node.network]
 events_buffer_len = 1000
 websocket_port = 9999
-trusted_node_addresses = ["/ip4/127.0.0.1/tcp/9998/ws"]
+node_addresses = ["/ip4/127.0.0.1/tcp/9998/ws"]
--- a/homestar-runtime/fixtures/settings.toml
+++ b/homestar-runtime/fixtures/settings.toml
@@ -6,4 +6,4 @@ process_collector_interval = 10
 [node.network]
 events_buffer_len = 1000
 websocket_port = 9999
-bootstrap_addresses = ["/ip4/127.0.0.1/tcp/9998/ws"]
+trusted_node_addresses = ["/ip4/127.0.0.1/tcp/9998/ws"]
--- a/homestar-runtime/fixtures/settings.toml
+++ b/homestar-runtime/fixtures/settings.toml
@@ -6,4 +6,4 @@ process_collector_interval = 10
 [node.network]
 events_buffer_len = 1000
 websocket_port = 9999
-bootstrap_addresses = ["/ip4/127.0.0.1/tcp/9998/ws/"]
+bootstrap_addresses = ["/ip4/127.0.0.1/tcp/9998/ws"]
--- a/homestar-runtime/src/event_handler.rs
+++ b/homestar-runtime/src/event_handler.rs
@@ -57,6 +57,7 @@ pub(crate) struct EventHandler<DB: Database> {
     sender: Arc<mpsc::Sender<Event>>,
     receiver: mpsc::Receiver<Event>,
     query_senders: FnvHashMap<QueryId, (RequestResponseKey, P2PSender)>,
+    connected_peers: FnvHashMap<PeerId, ConnectedPoint>,
     request_response_senders: FnvHashMap<RequestId, (RequestResponseKey, P2PSender)>,
     ws_msg_sender: ws::Notifier,
 }
@@ -104,6 +105,7 @@ where
             receiver,
             query_senders: FnvHashMap::default(),
             request_response_senders: FnvHashMap::default(),
+            connected_peers: FnvHashMap::default(),
             ws_msg_sender,
         }
     }
--- a/homestar-runtime/src/event_handler.rs
+++ b/homestar-runtime/src/event_handler.rs
@@ -10,7 +10,10 @@ use crate::{
 use anyhow::Result;
 use async_trait::async_trait;
 use fnv::FnvHashMap;
-use libp2p::{futures::StreamExt, kad::QueryId, request_response::RequestId, swarm::Swarm};
+use libp2p::{
+    core::ConnectedPoint, futures::StreamExt, kad::QueryId, request_response::RequestId,
+    swarm::Swarm, PeerId,
+};
 use std::{sync::Arc, time::Duration};
 use swarm_event::ResponseEvent;
 use tokio::{select, sync::mpsc};
@@ -48,6 +51,7 @@ pub(crate) struct EventHandler<DB: Database> {
     sender: Arc<mpsc::Sender<Event>>,
     receiver: mpsc::Receiver<Event>,
     query_senders: FnvHashMap<QueryId, (RequestResponseKey, P2PSender)>,
+    connected_peers: FnvHashMap<PeerId, ConnectedPoint>,
     request_response_senders: FnvHashMap<RequestId, (RequestResponseKey, P2PSender)>,
 }
 
@@ -71,6 +75,7 @@ where
             sender: Arc::new(sender),
             receiver,
             query_senders: FnvHashMap::default(),
+            connected_peers: FnvHashMap::default(),
             request_response_senders: FnvHashMap::default(),
         }
     }
--- a/homestar-runtime/src/event_handler/swarm_event.rs
+++ b/homestar-runtime/src/event_handler/swarm_event.rs
@@ -92,6 +92,8 @@ async fn handle_swarm_event<THandlerErr: fmt::Debug + Send, DB: Database>(
     event_handler: &mut EventHandler<DB>,
 ) {
     match event {
+        // TODO: add identify for adding compatable kademlia nodes.
+        // TODO: use kademlia to discover new gossip nodes.
         SwarmEvent::Behaviour(ComposedEvent::Gossipsub(gossip_event)) => match *gossip_event {
             gossipsub::Event::Message {
                 message,
@@ -379,16 +381,6 @@ async fn handle_swarm_event<THandlerErr: fmt::Debug + Send, DB: Database>(
         SwarmEvent::ConnectionEstablished {
             peer_id, endpoint, ..
         } => {
-            let behavior = event_handler.swarm.behaviour_mut();
-            // only listener addresses should be added to the routing table.
-            // TODO: add identify here to discover more listen addresses and add those.
-            if endpoint.is_listener() {
-                // ignores if the peer failed or not to be added in the routing table.
-                behavior
-                    .kademlia
-                    .add_address(&peer_id, endpoint.get_remote_address().clone());
-            }
-
             // add peer to connected peers list
             event_handler.connected_peers.insert(peer_id, endpoint);
         }
--- a/homestar-runtime/src/event_handler/swarm_event.rs
+++ b/homestar-runtime/src/event_handler/swarm_event.rs
@@ -376,6 +376,26 @@ async fn handle_swarm_event<THandlerErr: fmt::Debug + Send, DB: Database>(
             );
         }
         SwarmEvent::IncomingConnection { .. } => {}
+        SwarmEvent::ConnectionEstablished {
+            peer_id, endpoint, ..
+        } => {
+            let behavior = event_handler.swarm.behaviour_mut();
+            // only listener addresses should be added to the routing table.
+            // TODO: add identify here to discover more listen addresses and add those.
+            if endpoint.is_listener() {
+                // ignores if the peer failed or not to be added in the routing table.
+                behavior
+                    .kademlia
+                    .add_address(&peer_id, endpoint.get_remote_address().clone());
+            }
+
+            // add peer to connected peers list
+            event_handler.connected_peers.insert(peer_id, endpoint);
+        }
+        SwarmEvent::ConnectionClosed { peer_id, cause, .. } => {
+            info!("peer connection closed {peer_id}, cause: {cause:?}");
+            event_handler.connected_peers.remove_entry(&peer_id);
+        }
         _ => {}
     }
 }
--- a/homestar-runtime/src/network/swarm.rs
+++ b/homestar-runtime/src/network/swarm.rs
@@ -66,21 +66,20 @@ pub(crate) async fn new(settings: &settings::Node) -> Result<Swarm<ComposedBehav
 }
 
 fn startup<T: NetworkBehaviour>(swarm: &mut Swarm<T>, settings: &settings::Network) -> Result<()> {
-    // Dial bootstrap nodes specified in settings. Failure here shouldn't halt node startup.
+    // Listen-on given address
+    swarm.listen_on(settings.listen_address.to_string().parse()?)?;
+
+    // Dial trusted nodes specified in settings. Failure here shouldn't halt node startup.
     for trusted_addr in &settings.trusted_node_addresses {
         swarm
             .dial(trusted_addr.clone())
             .map(|_| {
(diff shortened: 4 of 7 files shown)


homestar-runtime/src/settings.rs
```diff
@@ -98,6 +98,9 @@ pub struct Network {
     pub(crate) workflow_quorum: usize,
     /// Pubkey setup configuration
     pub(crate) keypair_config: PubkeyConfig,
+    /// Multiaddrs of the bootstrap nodes to connect to on startup
+    #[serde_as(as = "Vec<serde_with::DisplayFromStr>")]
+    pub(crate) bootstrap_addresses: Vec<libp2p::Multiaddr>,
```
zeeshanlakhani, line 103: maybe `bootstrap_peers`?  Addresses comes through on the type.

homestar-runtime/src/network/swarm.rs
```diff
@@ -57,6 +57,18 @@ pub(crate) async fn new(settings: &settings::Node) -> Result<Swarm<ComposedBehav
     // Listen-on given address
     swarm.listen_on(settings.network.listen_address.to_string().parse()?)?;
 
+    // Dial bootstrap nodes specified in settings. Failure here shouldn't halt node startup.
+    for bootstrap_addr in &settings.network.bootstrap_addresses {
```
zeeshanlakhani, line 61: minor, but let's maybe hide this into an impl function and call it `fn bootstrap`? This may include the subscribe and listen steps.

homestar-runtime/src/network/swarm.rs
```diff
@@ -57,6 +57,18 @@ pub(crate) async fn new(settings: &settings::Node) -> Result<Swarm<ComposedBehav
...
+        swarm
+            .dial(bootstrap_addr.clone())
+            .map(|_| {
+                tracing::info!("Successfully dialed configured bootstrap node {bootstrap_addr}")
```
zeeshanlakhani, line 65: minor: have been importing info!/warn! directly.

homestar-runtime/src/network/swarm.rs
```diff
@@ -57,6 +57,18 @@ pub(crate) async fn new(settings: &settings::Node) -> Result<Swarm<ComposedBehav
     // Listen-on given address
     swarm.listen_on(settings.network.listen_address.to_string().parse()?)?;
 
+    // Dial bootstrap nodes specified in settings. Failure here shouldn't halt node startup.
+    for bootstrap_addr in &settings.network.bootstrap_addresses {
+        swarm
+            .dial(bootstrap_addr.clone())
```
zeeshanlakhani, line 63: so we dial now, but shouldn't we add bootstrap nodes, that succeed in dialing, as explicit peers?

mriise: I think this is reasonable for us to be doing at this point, but in this case I think we should probably rename it to "trusted_peers" or similar. Bootstrap nodes for larger (untrusted) networks normally are still included in peer scoring so that they can eventually leave a submesh after a new node builds it's own mesh up. I think it is best to avoid any potential confusion between bootstrap peers and (what this change will do) trusted peers.

zeeshanlakhani: yep, agreed on the naming convention(s).

homestar-runtime/src/network/swarm.rs
mriise: renaming to an even more generic `swarm_init`. eventally I'd like to put all startup things in it but we need to define an interface to work with networking.

homestar-runtime/src/settings.rs
mriise: as this is the name used as the key for the config file i would like to stay a bit explicit on the name.

mriise:
tests will probably fail at the moment since i havent setup nix properly on my laptop yet. needed to get this change uploaded though.

homestar-runtime/src/network/swarm.rs
```diff
@@ -65,6 +65,28 @@ pub(crate) async fn new(settings: &settings::Node) -> Result<Swarm<ComposedBehav
     Ok(swarm)
 }
 
+fn swarm_startup<T: NetworkBehaviour>(
```
zeeshanlakhani, line 68: minor: for consistency, since this is already in the swarm module, just `startup` should work. We'll essentially make `startup` a definition that's for any process that spins up via spawn.

zeeshanlakhani:
Ok, putting together some additional work here @mriise, to close this out:

* In `swarm_event`.rs, need to match on the `SwarmEvent::ConnectionEstablished` and add `kademlia.add_address` to add the peer id and peer address. I had implemented this earlier but wanted to tackle this separately again, as here, so 👍🏽. 
  - As part of this, we'll want the event handler to keep a local set (fnvhashset) of trusted peers that are actually "connected", vs those we configured. 
  - On dial, in the `startup` here, add addresses to the set and to kademlia. On, `SwarmEvent::OutgoingConnectionError` and `SwarmEvent::ConnectionClosed`, remove from the set. 
  - Currently, the set is just a "view" on trusted ones that are connected. Could be useful for query for now. 
* The explicit peer piece we discussed, as these are trusted, bootstrap nodes, since we configured them.

homestar-runtime/src/event_handler.rs
```diff
@@ -48,6 +51,7 @@ pub(crate) struct EventHandler<DB: Database> {
     sender: Arc<mpsc::Sender<Event>>,
     receiver: mpsc::Receiver<Event>,
     query_senders: FnvHashMap<QueryId, (RequestResponseKey, P2PSender)>,
+    connected_peers: FnvHashMap<PeerId, ConnectedPoint>,
```
zeeshanlakhani, line 54: 👍🏽

homestar-runtime/src/event_handler/swarm_event.rs
```diff
@@ -376,6 +376,26 @@ async fn handle_swarm_event<THandlerErr: fmt::Debug + Send, DB: Database>(
...
+            peer_id, endpoint, ..
+        } => {
+            let behavior = event_handler.swarm.behaviour_mut();
+            // only listener addresses should be added to the routing table.
+            // TODO: add identify here to discover more listen addresses and add those.
+            if endpoint.is_listener() {
+                // ignores if the peer failed or not to be added in the routing table.
+                behavior
+                    .kademlia
+                    .add_address(&peer_id, endpoint.get_remote_address().clone());
```
zeeshanlakhani, line 389: From looking at the libp2p code, we should always add an address to Kademlia upon dialing, as it may come back online later. If the endpoint is a dialer, then we can remove the address.

homestar-runtime/src/event_handler/swarm_event.rs
```diff
@@ -376,6 +376,26 @@ async fn handle_swarm_event<THandlerErr: fmt::Debug + Send, DB: Database>(
...
+            }
+
+            // add peer to connected peers list
+            event_handler.connected_peers.insert(peer_id, endpoint);
+        }
+        SwarmEvent::ConnectionClosed { peer_id, cause, .. } => {
+            info!("peer connection closed {peer_id}, cause: {cause:?}");
+            event_handler.connected_peers.remove_entry(&peer_id);
```
zeeshanlakhani, line 397: we should also remove an entry on the OutgoingConnectionError.

mriise: We only insert connections on successful connection anyways, so we shouldn't need to remove from the set on OutgoingConnectionError. Connections that are established and then err will route through ConnectionClosed.

homestar-runtime/src/event_handler/swarm_event.rs
mriise: not sure I understand what you mean here. Do you mean we should be adding an address when its a diler as well, in case identify gives us addresses the peer is listening on?

zeeshanlakhani: No, I mean, upon dialing, we should always add the address and then remove if the endpoint is a dialer. Essentially, we should do this in startup, not within the event.

homestar-runtime/src/event_handler/swarm_event.rs
zeeshanlakhani: This is me basing a bit of this on https://github.com/libp2p/rust-libp2p/blob/master/examples/file-sharing/src/network.rs#L319 and reading around. There's a feeling that maybe ConnectionClosed may not always fire or fire along with the error.

mriise: `OutgoingConnectionError` should only be triggered before `ConnectionEstablished` on an outgoing dial. Once we _have_ made a connection (and added them to connected peers) `ConnectionClosed` should be the only error type to trigger. Since we only add to connected peers on success, there shouldn't ever be a case where we need to remove a peer from `connected` in `OutgoingConnectionError`

zeeshanlakhani: @mriise yep, agreed, as per docs:

> A [ConnectionEstablished](https://docs.rs/libp2p/latest/libp2p/swarm/enum.SwarmEvent.html#variant.ConnectionEstablished) event is reported if the dialing attempt succeeds, otherwise a [OutgoingConnectionError](https://docs.rs/libp2p/latest/libp2p/swarm/enum.SwarmEvent.html#variant.OutgoingConnectionError) event is reported.

Though, interesting how the example doesn't necessarily adhere to this.

homestar-runtime/fixtures/settings.toml
```diff
@@ -6,3 +6,4 @@ process_collector_interval = 10
 [node.network]
 events_buffer_len = 1000
 websocket_port = 9999
+trusted_node_addresses = ["/ip4/127.0.0.1/tcp/9998/ws"]
```
zeeshanlakhani, line 9: for the setting, I rather name these `node addresses`. We'll assume their trusted, but a comment/doc saying these will be trusted would be good.

homestar-runtime/src/event_handler/swarm_event.rs
```diff
@@ -93,6 +93,8 @@ async fn handle_swarm_event<THandlerErr: fmt::Debug + Send, DB: Database>(
     event_handler: &mut EventHandler<DB>,
 ) {
     match event {
+        // TODO: add identify for adding compatable kademlia nodes.
+        // TODO: use kademlia to discover new gossip nodes.
```
zeeshanlakhani, line 97: this means rendezvous, right? I do know you're covering this in another branch nonetheless.

homestar-runtime/src/network/swarm.rs
```diff
@@ -53,15 +57,47 @@ pub(crate) async fn new(settings: &settings::Node) -> Result<Swarm<ComposedBehav
     )
     .build();
 
+    startup(&mut swarm, &settings.network)?;
+
+    Ok(swarm)
+}
+
+fn startup(swarm: &mut Swarm<ComposedBehaviour>, settings: &settings::Network) -> Result<()> {
```
zeeshanlakhani, line 65: Here's a possible cleanup of this bit, particularly b/c we don't use the result of `.ok()` and that inner match is unnecessary: 

```suggestion
fn startup(swarm: &mut Swarm<ComposedBehaviour>, settings: &settings::Network) -> Result<()> {
    // Listen-on given address
    swarm.listen_on(settings.listen_address.to_string().parse()?)?;

    // Dial trusted nodes specified in settings. Failure here shouldn't halt node startup.
    settings.trusted_node_addresses.iter().for_each(|addr| {
        let _ = swarm
            .dial(addr.clone())
            .map(|_| info!(addr=?addr, "successfully dialed configured node"))
            // log dial failure and continue
            .map_err(|e| error!(err=?e, "failed to dial node"));

        // add node to kademlia routing table
        if let Some(Protocol::P2p(peer_id)) = addr.iter().find(|i| matches!(i, Protocol::P2p(_))) {
            info!(addr=?addr, "added configured node to kademlia routing table");
            swarm
                .behaviour_mut()
                .kademlia
                .add_address(&peer_id, addr.clone());
        } else {
            warn!(addr=?addr, err="node address doesn't include peer ID",
                  "node not added to kademlia routing table")
        }
    });

    // join on `receipts` topic
    swarm
        .behaviour_mut()
        .gossip_subscribe(pubsub::RECEIPTS_TOPIC)?;

    Ok(())
}
```

I also trimmed the logs here b/c of assumptions on trusted/configured being the same thing.

zeeshanlakhani (commented):
@mriise some minor cleanup mentions here and a suggestion for `startup` cleanup.

homestar-runtime/src/event_handler/swarm_event.rs
```diff
@@ -368,6 +370,16 @@ async fn handle_swarm_event<THandlerErr: fmt::Debug + Send, DB: Database>(
             );
         }
         SwarmEvent::IncomingConnection { .. } => {}
+        SwarmEvent::ConnectionEstablished {
+            peer_id, endpoint, ..
+        } => {
+            // add peer to connected peers list
+            event_handler.connected_peers.insert(peer_id, endpoint);
```
zeeshanlakhani, line 377: I feel like we should add trusted nodes as explicit peers for gossipsub and remove them appropriately, right? I know we're implementing rendezvous too, so maybe that's redundant?, but trusted nodes discovered should be made explicit, right @mriise?

homestar-runtime/src/event_handler/swarm_event.rs
mriise: alltogether yes. though it is very helpful to have identify along with rendezvous to check if it supports it (same with kademlia, gossipsub, request-response, ect.)

homestar-runtime/src/event_handler/swarm_event.rs
mriise: the major issue is that explicit peers aren't counted towards the current mesh count of a topic. If all of the connected nodes are explicit then gossip won't let you send messages (since there is no-one counted in the mesh!).  That grows the minimum number of nodes in a network (using gossip) from just 2 up to explicit nodes + 1 non-explicit (since explicit peering agreements are supposed to be reciprocal).

homestar-runtime/src/network/swarm.rs
mriise: we dont use the result of dial either (just log) so the `.ok()` skips the compiler warning for not using a result, though really a `match` here makes more sense anyways.

zeeshanlakhani: @mriise Yeah, totally get that we don't use the result. Typically, `let _` is preferred over `.ok` as it's more explicit that we're not caring about it all, vs having an Option of None.

homestar-runtime/src/event_handler/swarm_event.rs
zeeshanlakhani: @mriise, I totally understand how they are not counted in the mesh, and the need for facilitating smaller networks is a great point. Nonetheless, it feels a bit odd to fully skip explicit peering agreement(s) altogether. Looking at [fleek's impl](https://github.com/fleek-network/ursa/blob/main/crates/ursa-network/src/service.rs#L406), they add explicit nodes for identity events that are in the same peer network, using the same protocol name for example. Nothing that prevents this iteration from going in, but something to still think about.

homestar-runtime/src/network/swarm.rs
zeeshanlakhani: Nonetheless, a `match` variant with the error logged is probably the way forward, to your point.

homestar-runtime/src/event_handler/swarm_event.rs
mriise: I think there is a place for explicit nodes, but for the moment we should be fine to ignore since it is more of an optimization.

zeeshanlakhani: yep, def can for now.

zeeshanlakhani:
@mriise I'm about to push a small reworking of when scheduling runs vs the ahead-of-time parsing. Let's try to get the last few bits lined up for this one and get it in. I know you've been pushing to the other branch as well :).

mriise:
ci shouldn't be failing like this...

zeeshanlakhani (approved):
Good work @mriise!

PR #232, opened 2023-08-08
refactor: for mdns, do not add explicit peer, just add to kademlia

zeeshanlakhani:
# Description

As per a convo with @mriise, we should not add mdns peers as explicit peers. Instead, add them to kademlia.

### DIFF
--- a/homestar-runtime/src/event_handler/swarm_event.rs
+++ b/homestar-runtime/src/event_handler/swarm_event.rs
@@ -341,23 +341,23 @@ async fn handle_swarm_event<THandlerErr: fmt::Debug + Send, DB: Database>(
             }
         },
         SwarmEvent::Behaviour(ComposedEvent::Mdns(mdns::Event::Discovered(list))) => {
-            for (peer_id, _multiaddr) in list {
+            for (peer_id, multiaddr) in list {
                 info!("mDNS discovered a new peer: {peer_id}");
                 event_handler
                     .swarm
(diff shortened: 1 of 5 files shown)


homestar-runtime/src/event_handler/swarm_event.rs
```diff
@@ -341,23 +341,23 @@ async fn handle_swarm_event<THandlerErr: fmt::Debug + Send, DB: Database>(
...
-            for (peer_id, _multiaddr) in list {
+            for (peer_id, multiaddr) in list {
                 info!("mDNS discover peer has expired: {peer_id}");
                 event_handler
                     .swarm
                     .behaviour_mut()
-                    .gossipsub
-                    .remove_explicit_peer(&peer_id);
+                    .kademlia
+                    .remove_address(&peer_id, &multiaddr);
```
mriise, line 360: mdns will expire too quickly at the moment so this might remove nodes too eagerly.

mriise (approved):

homestar-runtime/src/event_handler/swarm_event.rs
zeeshanlakhani: I'll wrap this around with something +1.

zeeshanlakhani: Ah, it's dependent on the mdns ttl. Will configure that.

Issue #214, opened 2023-07-28
Support rendezvous protocol for discovery


zeeshanlakhani:
# Summary

Add support for https://docs.libp2p.io/concepts/discovery-routing/rendezvous/. This can be combined with our use of pubsub/gossipsub, as peer can be discovered based on listening on the same topic. 

## Solution

Can be based on examples, e.g.: https://github.com/libp2p/rust-libp2p/tree/master/examples/rendezvous.

mriise:
Running rendezvous with the one provided by rust-libp2p doesn't run through pubsub, it is more of a fediverse-style where everyone can be their own server and everyone is their own client as well. IMO this approach should be more reliable since gossips isn't depending on itself to get its own peers.

zeeshanlakhani:
> Running rendezvous with the one provided by rust-libp2p doesn't run through pubsub, it is more of a fediverse-style where everyone can be their own server and everyone is their own client as well. IMO this approach should be more reliable since gossips isn't depending on itself to get its own peers.

The combined note, just comes from the docs more explicitly:

```For effective discovery, rendezvous can be combined with [libp2p publish/subscribe](https://docs.libp2p.io/concepts/pubsub/overview). At a basic level, rendezvous can bootstrap pubsub by discovering peers subscribed to a topic. The rendezvous would be responsible for publishing packets, subscribing, or unsubscribing from packet shapes.```

zeeshanlakhani:
Great work @mriise, thank you!

PR #236, opened 2023-08-14
feat: Rendezvous for peer discovery

mriise:
# Description

Add Rendezvous protocol for peer discovery. Also adds:
- Identify to ensure that the nodes we are talking to support it, along with the other protocols we use.
- Annouce addresses setting for addresses we know we are reachable through. These addresses are shared with rendezvous and kademlia. e.g. `/dnsaddr/homestar-node.mriise.net/tcp/9090`

## Link to issue

#214

## Type of change

Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Test plan (required)

TODO, need to figure out how we should our test network stack in general.


zeeshanlakhani:
@mriise this ready for review? should we move out of draft?

mriise:
> @mriise this ready for review? should we move out of draft?

ready for review!

homestar-runtime/src/event_handler/swarm_event.rs
```diff
@@ -93,14 +97,174 @@ async fn handle_swarm_event<THandlerErr: fmt::Debug + Send, DB: Database>(
...
+                identify::Event::Error { peer_id, error } => {
+                    error!(err=?error, peer_id=peer_id.to_string(), "error while attempting to identify the remote")
+                }
+                identify::Event::Sent { peer_id } => {
+                    debug!(peer_id = peer_id.to_string(), "sent identify info to peer")
+                }
+                identify::Event::Received { peer_id, info } => {
+                    debug!(peer_id=peer_id.to_string(), info=?info, "identify info received from peer");
+
+                    if !event_handler
```
zeeshanlakhani, line 112: minor: may be nice to chain this via [bool::then_some](https://doc.rust-lang.org/std/primitive.bool.html#method.then_some) or similar.

homestar-runtime/src/event_handler/swarm_event.rs
```diff
@@ -93,14 +97,174 @@ async fn handle_swarm_event<THandlerErr: fmt::Debug + Send, DB: Database>(
...
+                    // rendezvous
+                    // we are good to register self & discover with any node we contact. more peers = more better!
+                    if info.protocols.contains(&RENDEZVOUS_PROTOCOL_NAME) {
+                        // register self with remote
+                        if let Err(err) = behavior.rendezvous_client.register(
+                            Namespace::from_static(RENDEZVOUS_NAMESPACE),
+                            peer_id,
+                            None,
+                        ) {
+                            error!(
```
zeeshanlakhani, line 148: minor: error -> warn.

homestar-runtime/src/event_handler/swarm_event.rs
```diff
@@ -93,14 +97,174 @@ async fn handle_swarm_event<THandlerErr: fmt::Debug + Send, DB: Database>(
...
+                                err = format!("{err}"),
```
zeeshanlakhani, line 149: do we need to format! here? not just err = ?err (which will debug) or if this def impl's display, can just be err = err.

homestar-runtime/src/event_handler/swarm_event.rs
```diff
@@ -93,14 +97,174 @@ async fn handle_swarm_event<THandlerErr: fmt::Debug + Send, DB: Database>(
     event_handler: &mut EventHandler<DB>,
 ) {
     match event {
-        // TODO: add identify for adding compatable kademlia nodes.
         // TODO: use kademlia to discover new gossip nodes.
+        SwarmEvent::Behaviour(ComposedEvent::Identify(identify_event)) => {
+            match identify_event {
+                identify::Event::Error { peer_id, error } => {
+                    error!(err=?error, peer_id=peer_id.to_string(), "error while attempting to identify the remote")
```
zeeshanlakhani, line 104: warn.

zeeshanlakhani (commented):
Most of this is getting into good shape!

homestar-runtime/src/event_handler/swarm_event.rs
zeeshanlakhani, line 104: check: we can remove these `to_strings` if display is already implemented for peer_id?

homestar-runtime/src/event_handler/swarm_event.rs
mriise: err = err is not the `Display` impl, its the tracing `RegisterError` trait the error here is using thiserror which doesn't satisfy, so I used format to force it to use Display.

homestar-runtime/Cargo.toml
```diff
@@ -54,12 +67,31 @@ http-serde = "1.1"
 humantime = { workspace = true }
 indexmap = { version = "2.0", features = ["serde"] }
 ipfs-api = { version = "0.17", optional = true }
-ipfs-api-backend-hyper = { version = "0.6", features = ["with-builder", "with-send-sync"], optional = true }
+ipfs-api-backend-hyper = { version = "0.6", features = [
```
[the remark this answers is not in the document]
mriise: this was done by rustfmt automatically. I can try to undo though.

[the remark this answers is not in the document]
zeeshanlakhani: oh if @mriise says so, that's fine. It didn't seem to do it for me haha.

homestar-runtime/src/settings.rs
```diff
@@ -119,11 +119,14 @@ pub struct Network {
...
     pub(crate) workflow_quorum: usize,
-    /// Pubkey setup configuration
+    /// Pubkey setup configuration.
     pub(crate) keypair_config: PubkeyConfig,
     /// Multiaddrs of the trusted nodes to connect to on startup.
     #[serde_as(as = "Vec<serde_with::DisplayFromStr>")]
     pub(crate) node_addresses: Vec<libp2p::Multiaddr>,
+    /// Multiaddrs of the external addresses this node will announce to the network.
+    #[serde_as(as = "Vec<serde_with::DisplayFromStr>")]
+    pub(crate) announce_addresses: Vec<libp2p::Multiaddr>,
```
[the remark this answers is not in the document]
mriise: yes, this somehow got cut out when rebasing. added back now.

homestar-runtime/src/event_handler/swarm_event.rs
```diff
@@ -93,14 +97,174 @@ async fn handle_swarm_event<THandlerErr: fmt::Debug + Send, DB: Database>(
...
+                            .insert(rendezvous_node, cookie);
+                    } else {
+                        // don't add peers that aren't from our namespace
+                        warn!(peer_id=rendezvous_node.to_string(), namespace=?cookie.namespace(), "rendezvous peer gave records from an unexpected namespace");
+                        return;
+                    }
+                    // TODO: maybe do an async iter here or dial queue? this bit is IO heavy
+                    // dial discovered peers
+                    for registration in registrations {
+                        // TODO: do anything with ttl here?
```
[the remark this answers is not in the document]
mriise: I think we can mostly ignore ttl for now on the client.

homestar-runtime/src/event_handler/swarm_event.rs
mriise: I would prefer to keep as-is. it is more readable than a long chain wrapped with paren, e.g. 
```
(!long.chain(|| {})
    .of()
    .events())
    .then(|| {})
```

homestar-runtime/src/event_handler/swarm_event.rs
```diff
@@ -93,14 +97,174 @@ async fn handle_swarm_event<THandlerErr: fmt::Debug + Send, DB: Database>(
...
+                    debug!(peer_id=peer_id.to_string(), info=?info, "identify info received from peer");
+
+                    if !event_handler
+                        .swarm
+                        .external_addresses()
+                        .any(|addr| addr == &info.observed_addr)
+                    {
+                        // identify observed an external address that we weren't aware of
+                        // add it to the addresses we announce to other peers
+                        // TODO: this may not be a good thing to do if we are adding addresses from a private network that are not supposed to be public
```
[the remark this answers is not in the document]
mriise: Not in a entirely consistent manner. We can use the IANA definition of a local IP with https://doc.rust-lang.org/std/net/struct.Ipv4Addr.html#method.is_global to guess correctly most of the time though.

homestar-runtime/src/event_handler/swarm_event.rs
zeeshanlakhani: Yeah. That's fine for now. I think there's room for some general cleanup in this section, but we can refactor later for sure.

homestar-runtime/src/network/swarm.rs
```diff
@@ -42,7 +45,7 @@ pub(crate) async fn new(settings: &settings::Node) -> Result<Swarm<ComposedBehav
     let mut swarm = SwarmBuilder::with_tokio_executor(
         transport,
         ComposedBehaviour {
-            gossipsub: pubsub::new(keypair, settings)?,
+            gossipsub: pubsub::new(keypair.clone(), settings)?,
```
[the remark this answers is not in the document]
mriise: this is a one cost on load, so I wouldn't worry about it.

homestar-runtime/src/event_handler/swarm_event.rs
```diff
@@ -93,14 +97,174 @@ async fn handle_swarm_event<THandlerErr: fmt::Debug + Send, DB: Database>(
...
+                    if cookie.namespace() == Some(&Namespace::from_static(RENDEZVOUS_NAMESPACE)) {
+                        event_handler
+                            .rendezvous_cookies
+                            .insert(rendezvous_node, cookie);
+                    } else {
+                        // don't add peers that aren't from our namespace
+                        warn!(peer_id=rendezvous_node.to_string(), namespace=?cookie.namespace(), "rendezvous peer gave records from an unexpected namespace");
+                        return;
+                    }
+                    // TODO: maybe do an async iter here or dial queue? this bit is IO heavy
```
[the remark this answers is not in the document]
mriise: digging into rust-libp2p a bit more and it turns out we dont need to do this at all! they already take care of managing a pool of outgoing connections, so the `dial` function itself isn't IO heavy!

homestar-runtime/src/event_handler/swarm_event.rs
```diff
@@ -93,14 +97,178 @@ async fn handle_swarm_event<THandlerErr: fmt::Debug + Send, DB: Database>(
...
+                            .insert(rendezvous_node, cookie);
+
+                        // dial discovered peers
+                        for registration in registrations {
+                            // TODO: do anything with ttl here?
+                            let opts = DialOpts::peer_id(registration.record.peer_id())
+                                .addresses(registration.record.addresses().to_vec())
+                                .condition(libp2p::swarm::dial_opts::PeerCondition::Disconnected)
+                                .build();
+                            // TODO: we might be dialing too many peers here. Add settings to configure when we stop dialing new peers
```
mriise, line 197: Important to note, since rendezvous can give us hundreds of nodes and without limits we will dial _all_ of them, which we then will identify and dial all of those, leading to exponentially more dials until we try to dial all the nodes that use homestar. We are only working with small networks atm so this isnt horrible yet, and this PR is getting fairly beefy already so imo should be in a new PR.

zeeshanlakhani: @mriise makes sense. Can you open an enhancement issue to that we create some thresholding around this?

zeeshanlakhani:
@mriise will re-review updates today.

homestar-runtime/Cargo.toml
```diff
@@ -3,7 +3,12 @@ name = "homestar-runtime"
 version = "0.1.0"
 description = "Homestar CLI"
 keywords = ["ipfs", "workflows", "ipld", "ipvm"]
-categories = ["workflow-engines", "distributed-systems", "runtimes", "networking"]
+categories = [
```
zeeshanlakhani, line 6: FYI, did look into this and while cargo fmt doesn't do this officially, seems like there are extensions for vscode. I'm bringing-in one on my branch, called taplo that will take care of this locally, at least.

zeeshanlakhani: Going to resolve this, as we'll use taplo moving forward.

zeeshanlakhani (commented):
@mriise Things are looking good! Just one question, really.

homestar-runtime/src/event_handler/swarm_event.rs
```diff
@@ -93,14 +97,178 @@ async fn handle_swarm_event<THandlerErr: fmt::Debug + Send, DB: Database>(
     event_handler: &mut EventHandler<DB>,
 ) {
     match event {
-        // TODO: add identify for adding compatable kademlia nodes.
-        // TODO: use kademlia to discover new gossip nodes.
+        SwarmEvent::Behaviour(ComposedEvent::Identify(identify_event)) => {
+            match identify_event {
+                identify::Event::Error { peer_id, error } => {
```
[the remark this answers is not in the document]
zeeshanlakhani: @'ing @mriise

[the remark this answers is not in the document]
mriise: Not thats easily matchable if any. Ideally we would do some verification with peers to ensure that we are adding a valid address, but thats not quite reliable. 

I also realized after thinking about it that this might be an issue of limits as well since we could add a whole bunch of addresses that might not be valid. 

If we want to be safe the easiest thing to do is not auto-add new external addresses, only trusting addresses we get from settings. I am more on that side of things at the moment if you are ok with that option as well.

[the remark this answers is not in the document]
zeeshanlakhani: It's a good question. @mriise, when we add them as part of `identify:received` events, we should only be adding ones discovered using the protocol right, which should really only be homestar nodes. But, relatedly, we can't *always* trust a node even it advertises the same namespace. I think we can eventually have some way to validate things, as we won't be able to always configure nodes in settings. Maybe we can at least limit them in number for now?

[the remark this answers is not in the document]
zeeshanlakhani: As discussed, we'll work on a limit (via settings) and add a comment to revisit more on validation/verification of these peers.

homestar-runtime/src/event_handler.rs
```diff
@@ -59,7 +59,9 @@ pub(crate) struct EventHandler<DB: Database> {
     query_senders: FnvHashMap<QueryId, (RequestResponseKey, P2PSender)>,
     connected_peers: FnvHashMap<PeerId, ConnectedPoint>,
     request_response_senders: FnvHashMap<RequestId, (RequestResponseKey, P2PSender)>,
+    rendezvous_cookies: FnvHashMap<PeerId, Cookie>,
     ws_msg_sender: ws::Notifier,
+    external_address_limit: usize,
```
zeeshanlakhani, line 64: this setting should probably just be u32 or even u16 (always positive?). We can always use usize at the call site.

zeeshanlakhani, line 64: Need this on the non-websocket one.

PR #260, opened 2023-08-25
chore: Remove resolver setting

bgins:
# Description

- [x] Remove the `resolver` setting from the workspace `Cargo.toml`
- [x] Remove whitespace from the `enum-assoc` version

The `resolver` setting of `2` is implicit when using Rust 2021: https://doc.rust-lang.org/edition-guide/rust-2021/default-cargo-resolver.html

## Type of change

## Test plan (required)

Build and run the tests. Everything should work without failures.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -5,7 +5,6 @@ members = [
     "homestar-functions",
     "homestar-runtime",
     "homestar-wasm"]
-resolver = "2"
 
 [workspace.package]
 authors = [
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -5,7 +5,6 @@ members = [
     "homestar-functions",
     "homestar-runtime",
     "homestar-wasm"]
-resolver = "2"
```
zeeshanlakhani, line 8: tested this locally, and re-read the docs. +1.

PR #261, opened 2023-08-26
chore: Rename principle to principal

bgins:
# Description

- [x] Rename principle to principal

`rs-ucan` misspells "principal" as "principle". We can alias to name this correctly, and we should eventually update the misspelling in `rs-ucan`.

## Type of change

## Test plan (required)

Run the test suite. All tests should pass.


zeeshanlakhani (approved):
Haha. Yeah, I can't remember if I wanted to fix this up, but didn't, or it was just due to the upstream lib issue. 

Dope, thank you @bgins

PR #275, opened 2023-08-28
feat: Improve relay app node rendering and controls

bgins:
# Description

- [x] Only render nodes when previous task has completed
- [x] Remove task out-of-order error case
- [x] Disable a workflow play button when the other workflow is active
- [x] Prevent multiple runs of a workflow when it is active (spinner that replaces play button is not clickable)
- [x] Rename `task.active` to `task.selected` to better indicate that this is UI state
- [x] Remove receipt and instruction CID from completed task dropdown 

We had an issue where Firefox would receive receipts over the Websocket channel nearly simultaneously which resulted in out-of-order errors. This PR fixes that by delaying the rendering of nodes until each previous task has completed and removing the error case.

## Type of change

- [x] Bug fix (non-breaking change that fixes an issue)
- [x] Refactor 

## Test plan (required)

The play button should be disabled when the other workflow is running. The button should not handle click events when in a spinner state.

Test in Firefox. No errors should be present and workflows should run and render as expected.

### DIFF
--- a/examples/websocket-relay/relay-app/src/components/Controls.svelte
+++ b/examples/websocket-relay/relay-app/src/components/Controls.svelte
@@ -24,10 +24,10 @@
     const { task } = event.detail;
 
     tasks[task.workflowId] = tasks[task.workflowId].map((t) => {
-      t.active = false;
+      t.selected = false;
 
       if (t.id === task.id) {
-        t.active = true;
+        t.selected = true;
       }
 
       return t;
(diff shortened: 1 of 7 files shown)


zeeshanlakhani:
Adding some issues to fix:

- [ ] Make it work in Safari

zeeshanlakhani:
- [ ] Make it work in Safari
 - [ ] Smarter handling of order

zeeshanlakhani:
Prob worth a rebase :).

bgins:
>  Make it work in Safari

Haven't gotten this to work in Safari yet. We can take it on in a separate PR. Opened an issue to track it: https://github.com/ipvm-wg/homestar/issues/279

zeeshanlakhani (approved):

PR #294, opened 2023-09-05
fix: Remove top-level await uses

### LINKED ISSUE #279: Example relay app does not load in Safari
# Summary

## Problem

The example Websocket relay app does not load in Safari.

### Impact

Users are not able to test the app in Safari. Minimal impact because they can use Chrome or Firefox instead.

## Solution

Not clear.

# Detail

**Describe the bug**

When running the app, Safari displays a blank page and the error `Unhandled Promise Rejection: ReferenceError: Cannot access uninitialized variable.`

**To Reproduce**

1. Run the app with `npm run --prefix relay-app dev`
2. Navigate to `localhost:5173`
3. Open the console and see the error

**Expected behavior**

The page should load and display the app.

**Desktop (please complete the following information):**

 - OS: macOS Ventura 13.5
 - Browser Safari Version 16.6 (18615.3.12.11.2)

**Additional Information**

There are similar reports in this SvelteKit issue: https://github.com/sveltejs/kit/issues/2889, but none of the suggested resolutions have worked out in our case.

bgins:
# Description

This pull request fixes the following bugs:

- [x] #279 

This PR removes uses of top-level await which are not properly supported by Safari.

## Link to issue

Closes #279

## Type of change

- [x] Bug fix (non-breaking change that fixes an issue)

## Test plan (required)

Run the app in Safari. The page should load and the app should run.

### DIFF
--- a/examples/websocket-relay/README.md
+++ b/examples/websocket-relay/README.md
@@ -30,6 +30,7 @@ processing workflows that chain inputs and outputs using
 
    https://www.loom.com/share/b0f882adc2ea45709d1f3031b5e61e92?sid=29cb403e-c666-4753-82f5-e35bbb710151
 
+Note that IPFS may attempt to upgrade to a new version and produce an error after the update. Delete the `tmp/.ipfs/` directory and restart to reset the IPFS repo state.
 
 [ipfs]: https://ipfs.tech/
 [ipld]: https://ipld.io/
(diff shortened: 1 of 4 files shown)


zeeshanlakhani (approved):
Looks good. Will test this in Safari in the more immediate morning.

Issue #131, opened 2023-05-03
Networking: Multi-node libp2p testing :gear:


zeeshanlakhani:
# Summary

Test swarm, gossip, mdns, node-to-node connections, and event handling. 

Where to steal from: https://github.com/libp2p/rust-libp2p/blob/master/protocols/kad/src/behaviour/test.rs, and poss

### Goals

- [x] Test receipt gossip and DB update
- [x] Test receipt use over network (from dialed peer)
- [x] Test rendezvous setup (3 nodes)
- [X] Test MDNS dialing
- [x] Test nodes added/removed from DHT
- [X] Test Node Dialing

ETA: 2023-11-10

PR #335, opened 2023-09-27
test: Add network tests

bgins:
# Description

This pull request implements the following features:

- [x] Add RPC listens on address test
- [x] Add Websocket listens on address test
- [x] Add libp2p generates peer ID test
- [x] Add libp2p known peers connect test
- [x] Add libp2p peers connect after mDNS discovery test
- [x] Add `console_subscriber_port` setting
- [x] Update swarm event logs so `peer_id` is always the first label (we test logs assuming so)
- [x] Update logs for swarm peer connection closed event to debug and move `peer_id` to a label
- [x] Add `kill_homestar` function (to abstract out some repeated code)
- [x] Turn off tokio console when running tests
- [x] Reorganize integration test fixture ports into a sparse sequence across fixtures (gives us room to add tests, but keeps them somewhat in order).

## Link to issue

Implements #131

## Type of change

## Test plan (required)

We are adding a test to check on existing functionality.


homestar-runtime/tests/utils.rs
```diff
@@ -85,9 +88,31 @@ pub(crate) fn stop_all_bins() -> Result<()> {
...
+/// Retrieve process output.
+pub(crate) fn retrieve_output(proc: Child) -> String {
+    let output = proc.wait_with_output().expect("failed to wait on child");
+    let plain_stdout_bytes = strip_ansi_escapes::strip(output.stdout);
+    String::from_utf8(plain_stdout_bytes).unwrap()
+}
+
+/// Kill the Homestar server/binary
+pub(crate) fn kill_homestar(mut homestar_proc: Child) -> Child {
+    if let Ok(None) = homestar_proc.try_wait() {
```
zeeshanlakhani, line 100: without the extra stop, I wonder if there's an else or other cases to check here?

homestar-runtime/tests/network.rs
```diff
@@ -0,0 +1,172 @@
...
+        predicate::str::contains("peer_id=12D3KooWBYAug7e9eE7z1z1jaYjfVQCfRy8r3keVYD8jdsEGZHr")
+            .eval(stdout.as_str())
+    );
+
+    Ok(())
+}
+
+#[test]
+#[file_serial]
+fn test_libp2p_listens_on_address() -> Result<()> {
```
zeeshanlakhani, line 47: serial

homestar-runtime/tests/network.rs
```diff
@@ -0,0 +1,172 @@
...
+    path::PathBuf,
+    process::{Command, Stdio},
+};
+
+#[allow(dead_code)]
+static BIN: Lazy<PathBuf> = Lazy::new(|| assert_cmd::cargo::cargo_bin(BIN_NAME));
+
+#[test]
+#[file_serial]
+fn test_libp2p_generates_peer_id() -> Result<()> {
```
zeeshanlakhani, line 16: serial

homestar-runtime/tests/network.rs
bgins: Ah right, thanks 🙏

homestar-runtime/tests/utils.rs
bgins: My understanding is that if `try_wait` returns `Ok(Some(status))`, then the child has already exited so we should be safe without an else condition. 

In cases where we don't call `stop_homestar`, we will enter this condition every time and attempt to kill the process after a one second timeout. We probably don't want the timeout in this case. I think it was originally intended for cases where you call `stop_homestar` first and want to wait for that to resolve.

We could call `stop_homestar` every time, but I'm not certain what will happen in cases where we have multiple nodes running. Will `stop_homestar` stop them all? Alternatively, we could parameterize the wait duration to `kill_homestar` and kill right away where it makes sense.

zeeshanlakhani: @bgins yep. this was a bit before we found the issue that you fixed :).

homestar-runtime/tests/network.rs
```diff
@@ -0,0 +1,183 @@
...
+        .arg("--db")
+        .arg("homestar1.db")
+        .stdout(Stdio::piped())
+        .spawn()
+        .unwrap();
+
+    let dead_proc = kill_homestar(homestar_proc);
+    let stdout = retrieve_output(dead_proc);
+
+    assert_eq!(
```
zeeshanlakhani, line 65: don't need eq, just assert! on the predicate

homestar-runtime/tests/test_node1/config/settings.toml
```diff
@@ -1,13 +1,18 @@
...
-node_addresses = ["/ip4/127.0.0.1/tcp/9092/ws"]
+listen_address = "/ip4/127.0.0.1/tcp/7000"
+node_addresses = [
+  "/ip4/127.0.0.1/tcp/7001",
+  "/ip4/127.0.0.1/tcp/9092/ws"
+]
 
 [node.network.keypair_config]
-existing = { key_type = "ed25519", path = "../../fixtures/__testkey_ed25519.pem" }
+random_seed = { key_type = "ed25519", seed = "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA=" }
```
zeeshanlakhani, line 18: we should move to using an actual pem file for all the multi-node ones.

homestar-runtime/tests/utils.rs
bgins: Added a timeout parameter to `kill_homestar` to give us flexibility on how long we wait.

homestar-runtime/tests/utils.rs
```diff
@@ -58,7 +61,7 @@ pub(crate) fn stop_homestar() -> Result<()> {
     Ok(())
 }
 
-/// Stop the IPFS daemon.
+/// Stop the IPFS binary.
```
zeeshanlakhani, line 64: we should probably force remove this lock. 

```Error: cannot acquire lock: creation of lock C:\Users\runneradmin\.ipfs\repo.lock failed: The process cannot access the file because it is being used by another process.```

Comment (bgins) [reply homestar-runtime/tests/utils.rs:64]: Coming in https://github.com/ipvm-wg/homestar/pull/359. Thanks! 🙂

File: homestar-runtime/tests/network.rs [hunk 7]
```diff
@@ -0,0 +1,255 @@
...
+use once_cell::sync::Lazy;
+use predicates::prelude::*;
+use serial_test::file_serial;
+use std::{
+    path::PathBuf,
+    process::{Command, Stdio},
+    time::Duration,
+};
+
+#[allow(dead_code)]
```
zeeshanlakhani, line 12: minor: probably can be removed (dead_code)

zeeshanlakhani:
@bgins on the coverage test, the port I configured seems to be related to something on GH runs where it wants to sometimes kill it. I'm experimenting with some diff ports there.

homestar-runtime/tests/network.rs
bgins: Yep! Changed this on all the tests. 👍

zeeshanlakhani (approved):

Issue #350, opened 2023-10-04
Use local IPFS URIs rather than global gateway


bmann:
The workflows in the websocket-relay example app have rsc fields that are URIs to the global IPFS gateway

`"rsc": "https://ipfs.io/ipfs/bafybeiabbxwf2vn4j3zm7bbojr6rt6k7o6cg6xcbhqkllubmsnvocpv7y4"`

As I understand it, both the wasm module and synthcat are in the repo, and are `ipfs add`-ed as part of running the example. We should use either an IPFS URI or Localhost gateway path.

e.g.

`ipfs://bafybeiabbxwf2vn4j3zm7bbojr6rt6k7o6cg6xcbhqkllubmsnvocpv7y4`

or 

`http://localhost:8080/ipfs/bafybeiabbxwf2vn4j3zm7bbojr6rt6k7o6cg6xcbhqkllubmsnvocpv7y4`

That hash is the wasm, and when I use this path while running the demo, it downloads a Wasm file, as `downloads.wasm`. Yay! (But also: don't think one can do anything interesting with that)

Using the synthcat hash, redirects to an origin hash subdomain version `http://bafybeiejevluvtoevgk66plh5t6xiy3ikyuuxg3vgofuvpeckb6eadresm.ipfs.localhost:8080/` and shows me synthcat in all their glory.

Ideally, this example should run without being connected to the Internet.

Issue #348, opened 2023-10-04
Running websocket-relay doesn't work if you already have IPFS running


bmann:
I have IPFS Desktop running which runs a kubo instance locally

If I have kubo running and run the `cargo run -- start -c config/settings.toml` command as listed in [examples/websocket-relay](https://github.com/ipvm-wg/homestar/blob/main/examples/README.md) it doesn't work.

Recommendation:
* if we're going to puppet kubo like this we should use the not-default-port
* future: config option for people to run a binary of their choice / port of their choice (and even with examples, I'd rather us start leaning into iroh here)

zeeshanlakhani:
Closed by #352

PR #352, opened 2023-10-04
chore: README updates, example updates

zeeshanlakhani:
Fixes: https://github.com/ipvm-wg/homestar/issues/350 and https://github.com/ipvm-wg/homestar/issues/348


bgins (approved):
Looks good! I've tested this under a few conditions:

- Connected to the Internet 
  - While running an IPFS daemon and IPFS desktop
  - Without an IPFS daemon, the example app runs one up
- Offline
  - While running an IPFS daemon and IPFS desktop
  - Without an IPFS daemon, the example app runs one up

The example works in all of these cases.

PR #355, opened 2023-10-05
Add Node 18 LTS to Flake for example frontends

expede:
# Description

Websocket example app depends on Nodejs, and (from a side conversation) a minimum version. This PR adds Node 18 LTS to to the Nix Flake, and rebuilds the Websocket relay lockfile. This means that if you run `nix develop`, you will have a consistent version of Node on your `PATH`.

## Type of change

## Test plan (required)

`#worksonmymachine`?

## Screenshots/Screencaps

Screenshot 2023-10-04 at 23 11 57


zeeshanlakhani:
I think @bgins's fork had an issue on the version, which was fixed by the removal of svelte-lime. Nonetheless, good to include this min version anyway, in the flake.

zeeshanlakhani (approved):

expede:
> Nonetheless, good to include this min version anyway, in the flake.

I don't even have Node installed globally, which is definitely a weird Nix person thing 😆

PR #389, opened 2023-10-19
feat: Add Gossipsub config setting

### LINKED ISSUE #388: Add enable Gossipsub config setting
# Summary

## Problem

We run libp2p Gossipsub by default to share receipts amongst peers, but there are some circumstances where we don't want gossip.

### Impact

Testing can be challenging when we want to check that receipts were requested at execution time.

Some binary users may want to turn off Gossipsub, but don't have an option for it.

## Solution

Add a config to turn the Gossipsub behavior on or off.

bgins:
# Description

This PR implements the following features:

- [x] Add `enable_gossipsub` config
- [x] Add `PubSubError` with `InsufficientPeers`, `NotEnabled`, `Conversion`, and wrapped Gossipsub `SubscriptionError` and `PublishError` error  cases
- [x] Add runtime network `Error` with `PubSubError`
- [x] Handle Gossipsub on/off cases for `gossip_subscribe` and `gossip_publish`

## Link to issue

Closes #388

## Type of change

## Test plan (required)

All existing tests should pass. We will test Gossipsub on/off in upcoming multi-node workflow integration tests.


homestar-runtime/src/network/swarm.rs
```diff
@@ -265,30 +277,39 @@ pub(crate) struct ComposedBehaviour {
...
-        let topic = gossipsub::IdentTopic::new(topic);
-        self.gossipsub.subscribe(&topic)
+    pub(crate) fn gossip_subscribe(&mut self, topic: &str) -> Result<bool> {
+        if let Some(gossipsub) = self.gossipsub.as_mut() {
+            let topic = gossipsub::IdentTopic::new(topic);
+            let subscribed = gossipsub.subscribe(&topic)?;
+
+            Ok(subscribed)
+        } else {
+            Err(anyhow!(NetworkError::PubSubError(PubSubError::NotEnabled)))
```
zeeshanlakhani, line 287: this doesn't have to be anyhow'ed. It could just return NetworkEerror::PubSubError, no? The `?` at the call site should still work, like before.

homestar-runtime/src/network/swarm.rs
```diff
@@ -265,30 +277,39 @@ pub(crate) struct ComposedBehaviour {
...
+            if gossipsub
+                .mesh_peers(&TopicHash::from_raw(topic))
+                .peekable()
+                .peek()
+                .is_some()
+            {
+                let msg_id = gossipsub.publish(id_topic, msg_bytes)?;
+                Ok(msg_id)
+            } else {
+                Err(anyhow!(
```
zeeshanlakhani, line 307: same as above (and this can become another pubsub error?)

zeeshanlakhani (commented):
A couple of minor questions if we need anyhow or can rely on our errors (and get the `?`) at the call site (which we should?). 

Otherwise 👍🏽

zeeshanlakhani (approved):

PR #399, opened 2023-10-24
refactor: Improve rendezvous protocol usage

bgins:
# Description

This pull request implements the following features:

- [x] Add event handler cache implementation
- [x] Renew rendezvous registration at expiration
- [x] Re-discover from rendezvous servers on a set interval
- [x] Renew discovered peer registrations when they expire
- [x] Remove behavior where we add other peers to `external_addresses`
- [x] Add check to not dial ourselves on rendezvous discovery
- [x] Update dialing on rendezvous discovery to only dial peers we aren't already connected to
- [x] Add rendezvous server `Dialing`, `PeerRegistered`, and `RegistrationExpired` event debug logs
- [x] Separate `enable_rendezvous` config into `enable_rendezvous_server` and `enable_rendezvous_client` configs (server is opt-in, client is opt-out)
- [x] Add `max_connected_peers` config
- [x] Add `rendezvous_registration_ttl` and `rendezvous_discovery_interval` configs
- [x] Test rendezvous register, discover, and connect
- [x] Test disconnect after rendezvous connect
- [x] Test registration expires and re-registration on expiration
- [x] Test discovered registration expires and rediscovery attempted at expiration
- [x] Test rediscovery on rendezvous discovery interval
- [x] Add `extract_timestamps_where` and `count_lines_where` test utilities
- [x] Update libp2p deprecated Kademlia and SwarmBuilder interfaces
- [x] Update integration test fixtures to avoid overlapping ports

## Link to issue

Implements #131.

## Type of change

- [x] Comments have been added/updated

## Test plan (required)

We are adding tests to check existing functionality and a test to check registration renewals.


homestar-runtime/src/event_handler/swarm_event.rs
```diff
@@ -194,15 +195,25 @@ async fn handle_swarm_event<THandlerErr: fmt::Debug + Send, DB: Database>(
...
                             let opts = DialOpts::peer_id(registration.record.peer_id())
                                 .addresses(registration.record.addresses().to_vec())
                                 .condition(libp2p::swarm::dial_opts::PeerCondition::Disconnected)
                                 .build();
-                            // TODO: we might be dialing too many peers here. Add settings to configure when we stop dialing new peers
-                            if let Err(err) = event_handler.swarm.dial(opts) {
-                                warn!(peer_id=registration.record.peer_id().to_string(), err=?err, "failed to dial peer discovered through rendezvous")
+
+                            // Dial discovered peer if not at connected peers limit
+                            if event_handler.connected_peers.len() + index
```
zeeshanlakhani, line 206: minor: we can calculate the len() outside the loop, right?

zeeshanlakhani: Also, if the connected_peers list is already > than the limit, then we don't even need the loop.

bgins: We want to calculate inside the loop because we want to dial some discovered peers, but not all of them. 

For example, say our `connected_peers_limit` is 50 and we are currently connected to 40 peers. We receive a set of registrations that includes 20 peers. We can loop and dial the first 10 discovered peers, and ignore the remaining 10 peers.

Thinking about this some more, we can probably do better. We could add a check to not dial peers that we are already connected to. That way, we can maximize the set of registrations to establish the most connections.

zeeshanlakhani: Think we still want some of the other pre-checks, but I like the last idea to avoid the extra dials.

bgins: Updated to only dial peers we aren't already connected to. 

Had a moment when I thought we might be re-implementing the benefits we get from rendezvous cookies, but I think this is still valuable because we might dial peers outside of a rendezvous discovery context.

homestar-runtime/src/event_handler.rs
```diff
@@ -54,11 +57,16 @@ pub(crate) struct EventHandler<DB: Database> {
...
     swarm: Swarm<ComposedBehaviour>,
+    cache: Arc<Cache<String, CacheValue>>,
     sender: Arc<channel::AsyncBoundedChannelSender<Event>>,
     receiver: channel::AsyncBoundedChannelReceiver<Event>,
     query_senders: FnvHashMap<QueryId, (RequestResponseKey, Option<P2PSender>)>,
     connected_peers: FnvHashMap<PeerId, ConnectedPoint>,
+    connected_peers_limit: u32,
+    discovered_peers: FnvHashMap<PeerId, PeerDiscoveryInfo>,
     request_response_senders: FnvHashMap<RequestId, (RequestResponseKey, P2PSender)>,
+    rendezvous_registration_ttl: Duration,
```
zeeshanlakhani, line 68: @bgins thinking maybe we do have some substruct here: rendezvous, which has ttl and discovery interval and cookies? Trying to clean up this bit on length here. We may want the same for the two connected peers.

homestar-runtime/src/event_handler.rs
```diff
@@ -110,17 +123,23 @@ where
...
+        let sender = Arc::new(sender);
         Self {
             receipt_quorum: settings.network.receipt_quorum,
             workflow_quorum: settings.network.workflow_quorum,
             p2p_provider_timeout: settings.network.p2p_provider_timeout,
             db,
             swarm,
-            sender: Arc::new(sender),
+            cache: Arc::new(setup_cache(sender.clone())),
+            sender: sender.clone(),
```
zeeshanlakhani, line 134: did we need to arc outside and clone in here?

homestar-runtime/src/event_handler.rs
```diff
@@ -134,17 +153,23 @@ where
...
```
zeeshanlakhani, line 164: same as above?

homestar-runtime/src/event_handler.rs
```diff
@@ -134,17 +153,23 @@ where
...
             receiver,
             query_senders: FnvHashMap::default(),
             connected_peers: FnvHashMap::default(),
+            connected_peers_limit: settings.network.max_connected_peers,
+            discovered_peers: FnvHashMap::default(),
             request_response_senders: FnvHashMap::default(),
+            rendezvous_registration_ttl: settings.network.rendezvous_registration_ttl,
```
zeeshanlakhani, line 171: same as above.

homestar-runtime/src/event_handler.rs
```diff
@@ -189,15 +214,20 @@ where
                      swarm_event.handle_event(&mut self).await;
 
                 }
-
             }
+
+            // Poll cache for expired entries
+            self.cache.run_pending_tasks().await;
```
zeeshanlakhani, line 220: we have our spawn loop below, right? So this can go?

homestar-runtime/src/event_handler.rs
```diff
@@ -213,3 +243,13 @@ where
         }
     }
 }
+
+/// Poll cache for expired entries
+async fn poll_cache(cache: Arc<Cache<String, CacheValue>>) {
+    let mut interval = tokio::time::interval(Duration::from_secs(1));
```
zeeshanlakhani, line 249: let's make this configurable, maybe under monitoring?

homestar-runtime/src/event_handler/swarm_event.rs
```diff
@@ -187,26 +197,96 @@ async fn handle_swarm_event<THandlerErr: fmt::Debug + Send, DB: Database>(
...
+                            .filter(|registration| {
+                                !event_handler
+                                    .connected_peers
+                                    .contains_key(&registration.record.peer_id())
+                            })
+                            .collect();
+
+                        // Dial newly discovered peers
+                        for (index, registration) in new_registrations.iter().enumerate() {
+                            let our_registration = &registration.record.peer_id()
```
zeeshanlakhani, line 231: minor: our feels weird naming-wise. `self_registration`?

zeeshanlakhani (commented):
First pass. Need to look at the tests a little more, but this is looking great.

homestar-runtime/src/event_handler.rs
bgins: Ah yeah, missed updating this in the IPFS feature off case. Fixed. 👌

homestar-runtime/src/event_handler/swarm_event.rs
bgins: Yup that is better. Renamed it. 👍

homestar-runtime/src/event_handler.rs
bgins: Was able to simplify it to this:

https://github.com/ipvm-wg/homestar/blob/402a4121587bda974c27ec8c93dc86538aa61a06/homestar-runtime/src/event_handler.rs#L133-L134

homestar-runtime/src/event_handler.rs
bgins: Done! Also added `discovered_peers` to the rendezvous one because it is not relevant for peers discovered through mDNS. We may revisit if we use the proposed ambient discovery protocol in the future.

Was slightly unsure about the naming for the connected peers one. Went with this struct:

https://github.com/ipvm-wg/homestar/blob/1d4079b19884b9b130ac9a6274b96ed04f54720f/homestar-runtime/src/event_handler.rs#L104-L108

Which on the event handler is:

https://github.com/ipvm-wg/homestar/blob/1d4079b19884b9b130ac9a6274b96ed04f54720f/homestar-runtime/src/event_handler.rs#L64

And when used we have `event_handler.connections.peers` and `event_handler.connections.max_peers`. In the first one, having `connections` and `peers` seems a bit redundant? I might be overthinking it.

homestar-runtime/src/event_handler.rs
bgins: Done. Put it in the network config for now, but we can shuffle it around later.

PR #433, opened 2023-11-14
fix: make configuration run without -c (all defaults)

zeeshanlakhani:
Includes:

- also, add starter functions for default .pem/settings initialization
- error handling around retry, wasms
- ups the transport timeout


bgins (approved):
Looks good! Some nice DX improvements here. 💯

PR #435, opened 2023-11-14
refactor: static/dynamic nodeinfo and recv deadlines

zeeshanlakhani:


bgins (approved):
Looks good, let's bring it in. 👌

PR #437, opened 2023-11-15
fix: schedule + db conflict

zeeshanlakhani:


bgins (approved):
Tested these changes running workflows across a couple of nodes, gossiping receipts and with replayed workflows. Works well, found no issues.

Issue #491, opened 2024-01-05
Testing: integration test refactoring


zeeshanlakhani:
# Summary

As our codebase grows, we want to make working with integration tests easier!

## Components

- [x] Make serial, integration tests parallelized (integration-grouped) tests
- [x] Use inlined configuration over file(s) #485 
- [ ] ~~Set RPC/HTTP ports to 0 in test configurations and have the OS choose an unused port for all RPC/HTTP servers (via @QuinnWilton). *Note*: TCP listeners in Rust do this automatically with bind and using the TcpListenerStream -> https://github.com/hyperium/tonic/blob/master/tests/integration_tests/tests/timeout.rs#L79.~~ 
- [ ] ~~Use `:memory` DB configuration name in integration tests to remove the need for named setup/teardown. This should be isolated per test due to be running within a child proc.~~

zeeshanlakhani:
We've decided to punt on the last two checkboxes here. We may return to them in the future.

PR #528, opened 2024-01-23
port selection and test config generation macro

### LINKED ISSUE #485: Testing: Integration tests should use inlined configuration vs fixture/test-config files
# Summary

For our suite of integration tests, we've been relying on settings/configurations to separate ports, on/offs, and the like for nodes that get spun up in the tests so that there's no port conflict, etc.

As the test suite has grown, this has become too cumbersome. We need to move toward a model (or macro in this case!) where we can define the configuration inline, including randomization and port-checking for ports (which we already have utils for). When I say inline, this still means we write a file, but we should be able to specify the configuration in code, have that generate a file, and then clean up the file accordingly. 

## Components

- [ ] Create a `config!`, toml-driven macro that allows for specifying settings, including ports, and functions for things like random port(s). This macro should allow for the generation of different files within the same integration test, as we have multi-node tests. 
- [ ] Files that were dynamically created for the test should be easily accessible when starting up the node and then removed upon completion of the test on shutdown, automatically.

zeeshanlakhani:
# Description

@euamcg added a macro for inlining toml configs for tests. As we can see the amount of .toml configs is getting quite large. The macro allows us to write toml configs inside tests, create a fixture file, and discard it when the test is finished. 

This closes #485 and #491 (for now, as we may come back to some optimizations down the line related to the latter). 

## Link to issue

- [Issue#485](https://github.com/ipvm-wg/homestar/issues/485)
- [Issue#491](https://github.com/ipvm-wg/homestar/issues/491)

## Type of change

- [X] Refactor (non-breaking change that updates existing functionality)
- [X] This change requires a documentation update

Please delete options that are not relevant.

## Test plan (required)

All tests are passing.


homestar-runtime/src/main.rs
```diff
@@ -21,6 +21,7 @@ fn main() -> Result<()> {
             // Load settings first, so we can daemonize before starting the
             // runtime.
             let settings = if let Some(file) = runtime_config {
+                println!("Loading settings from file: {}", file.display());
```
bgins, line 24: Should this print line be left in?

homestar-runtime/tests/utils.rs
```diff
@@ -247,6 +261,45 @@ pub(crate) fn remove_db(name: &str) {
...
+            ))?,
+            db_path: PathBuf::from_str(&name)?,
+            listen_port: port_selector::random_free_port().ok_or(anyhow::anyhow!(
+                "failed to select free port for listen address"
+            ))?,
+        };
+
+        Ok(proc_info)
+    }
+}
```
bgins, line 293: Nice! Good stuff 💯

homestar-runtime/tests/utils.rs
```diff
@@ -291,3 +344,66 @@ pub(crate) trait TimeoutFutureExt<T>: Future<Output = T> + Sized {
...
+    }};
+    // For some finer control over the config.
+    ($name:expr, $toml:expr) => {{
+        let name = format!("tests/fixtures/{}.toml", $name);
+        let toml = $toml.parse::<toml::Table>().unwrap();
+        let test_config = $crate::utils::TestConfig::new(name.to_string(), toml);
+        test_config.save_fixture().unwrap();
+        test_config
+    }};
+}
```
bgins, line 409: 💖

bgins (approved):
Looks great! Nice to get rid of all those fixtures, loving the inline TOML. ❤️

PR #529, opened 2024-01-23
feat: redial `node_addresses` at an interval on connection close

### LINKED ISSUE #400: Networking: use event_handler cache to (re-)dial disconnected or connection-error node addresses
# Summary

Now that we have a mechanism to trigger events related to peer and swarm information within an event cache (which was added in #399), we should re-dial disconnected, trusted `node_addresses` and those that have initial connection errors. 

## Solution

Leverage our event cache in #399 to insert disconnected/error'ed connection node_addresses and return to them at an interval (leveraging expiry) to re-dial or reconnect. 

ETA: 2023-12-20

QuinnWilton:
This resolves #400.

### DIFF
--- a/homestar-runtime/src/event_handler/swarm_event.rs
+++ b/homestar-runtime/src/event_handler/swarm_event.rs
@@ -1157,6 +1157,32 @@ async fn handle_swarm_event<DB: Database>(
                     peer_id = peer_id.to_string(),
                     "removed peer from kademlia table"
                 );
+            } else {
+                debug!(
+                    subject = "libp2p.conn.closed",
+                    category = "handle_swarm_event",
+                    peer_id = peer_id.to_string(),
+                    "redialing trusted peer in {interval:?}",
+                    interval = event_handler.connections.dial_interval
+                );
+
+                // Dial peers again at dial interval
+                event_handler
+                    .cache
+                    .insert(
+                        format!("{}-dial", peer_id),
+                        CacheValue::new(
+                            event_handler.connections.dial_interval,
+                            HashMap::from([
+                                (
+                                    "on_expiration".to_string(),
+                                    CacheData::OnExpiration(cache::DispatchEvent::DialPeer),
+                                ),
+                                ("node".to_string(), CacheData::Peer(peer_id)),
+                            ]),
+                        ),
+                    )
+                    .await;
             }
 
             #[cfg(feature = "websocket-notify")]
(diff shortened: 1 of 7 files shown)


QuinnWilton:
Briefly chatted with Brian and we came up with an idea for writing an integration test for this, so I'll give that a try still, + also handle retrying when a redial fails to reestablish the connection.

homestar-runtime/tests/network/notification.rs
```diff
@@ -184,3 +184,385 @@ fn test_connection_notifications_integration() -> Result<()> {
...
+
+                if json["type"].as_str().unwrap() == "network:connectionEstablished" {
+                    break;
+                }
+            } else {
+                panic!("Node one did not redial node two in time.")
+            }
+        }
+
+        kill_homestar(proc_guard1.take(), None);
```
zeeshanlakhani, line 563: Is the plan to check logs too (with these being here)?

homestar-runtime/src/event_handler/swarm_event.rs
```diff
@@ -1182,6 +1208,50 @@ async fn handle_swarm_event<DB: Database>(
...
+            // Redial peer if in configured peers
+            if let Some(peer_id) = peer_id {
+                if event_handler.node_addresses.iter().any(|multiaddr| {
+                    if let Some(id) = multiaddr.peer_id() {
+                        id == peer_id
+                    } else {
+                        // TODO: We may want to check the multiadress without relying on
+                        // the peer ID. This would give more flexibility when configuring nodes.
+                        warn!(
+                            subject = "libp2p.conn.closed",
```
bgins, line 1220: We should update the subject here to `libp2p.outgoing.err`.

Also, maybe worth a helper function now that we are using this conditional check twice?

bgins (approved):
Looks good! ✨ Thanks for taking this one on. Left a comment for one small change.

homestar-runtime/tests/network/notification.rs
zeeshanlakhani: @QuinnWilton we can remove these.

homestar-runtime/src/event_handler/swarm_event.rs
QuinnWilton: Thanks! I updated the subject, but haven't extracted to a helper function yet, because of the logging that happens inside of the conditional check that needs to be different at both call sites. If you think it's worthwhile, we can just pass the subject as a parameter though.

bgins: Yeah, that does make it a bit awkward because it's not just a conditional check. I'm alright leaving it as is. 👌

homestar-runtime/tests/network/notification.rs
```diff
@@ -184,3 +184,382 @@ fn test_connection_notifications_integration() -> Result<()> {
...
```
zeeshanlakhani, line 560: @QuinnWilton I don't think we need the kills at all b/c they will be killed automatically on drop, due to the guard.

homestar-runtime/tests/network/notification.rs
```diff
@@ -184,3 +184,382 @@ fn test_connection_notifications_integration() -> Result<()> {
 
     Ok(())
 }
+
+#[test]
+fn test_libp2p_redial_on_connection_closed_integration() -> Result<()> {
+    let proc_info1 = ProcInfo::new().unwrap();
```
zeeshanlakhani, line 190: windows has -D warnings on, so it's looking for the `_proc_info1` etc.

homestar-runtime/src/event_handler/swarm_event.rs
```diff
@@ -1182,6 +1208,50 @@ async fn handle_swarm_event<DB: Database>(
                   "outgoing connection error"
             );
 
+            // Redial peer if in configured peers
+            if let Some(peer_id) = peer_id {
+                if event_handler.node_addresses.iter().any(|multiaddr| {
+                    if let Some(id) = multiaddr.peer_id() {
+                        id == peer_id
+                    } else {
+                        // TODO: We may want to check the multiadress without relying on
```
zeeshanlakhani, line 1217: good future todo.

zeeshanlakhani (approved):
A minor thing/warning thing or two on warnings/tests. Otherwise 👍🏽, great work @QuinnWilton.

homestar-runtime/tests/network/notification.rs
QuinnWilton: Ah whoops, I forgot there was a second test to remove these from! Fixing it now.

expede:
Just testing the webhook — don't mind me

Issue #422, opened 2023-11-10
Generate keypair when not provided


bgins:
# Summary

## Problem

Homestar requires users to generate a keypair or use a randomly generated one.

### Impact

Manually generating a keypair is inconvenient. Randomly generated keypairs result in a new peer ID on each run, which makes it challenging to configure network discovery and connections.

## Solution

We will want a default keypair location. When a keypair path is not provided, check the default location. If no keypair is available at the default location, generate one and persist it.

zeeshanlakhani:
Closed with https://github.com/ipvm-wg/homestar/pull/582.

PR #582, opened 2024-02-21
Load config from well-known locations + add `init` command for initializing config

QuinnWilton:
# Description

This adds the start of an `init` command for configuring Homestar from the CLI. When run interactively, it currently provides a wizard for configuring the keys used by homestar, however it can also be run non-interactively using values configured through CLI flags.

It also includes allowances for outputting a generated configuration over stdout, and the start of some patterns for handling some common CLI knobs, like `--no-input`, `--quiet`, and `--force`, though these are only likely to see more use as we expose more configuration options to this interactive wizard.

There's still a lot of polish that can be done around how we handle error reporting and interrupts, like control+c, during interactive usage, but this is a start that should make it a lot easier to begin expanding the configuration options we expose to the user.

## Link to issue

[Please add a link to any relevant issues/tickets.](https://github.com/ipvm-wg/homestar/issues/422)

## Screenshots/Screencaps

image
image
image
image
image


homestar-runtime/Cargo.toml
```diff
@@ -38,28 +38,29 @@ atomic_refcell = { workspace = true }
...
-  "help",
-  "env",
-  "std",
-  "usage",
+    "derive",
+    "color",
+    "help",
+    "env",
+    "std",
+    "usage",
```
bgins, line 46: It looks like this file may been formatted with four spaces of indentation. Is this a proposal to move to that instead of two?

zeeshanlakhani: @bgins we should let the toml checker thing (taplo?) handle this. Maybe it is?

bgins: Cargo manifest conventions suggest it should be the same indentation as Rust code: https://doc.rust-lang.org/stable/style-guide/cargo.html, so four is probably better.

QuinnWilton: Shoot, yeah, this was my editor autoformatting using `taplo`. I can split the autoformatting out into a separate commit if you'd like.

zeeshanlakhani: @QuinnWilton it's fine. As long as it as taplo for all of us!

homestar-runtime/src/main.rs
```diff
@@ -13,6 +15,36 @@ fn main() -> Result<()> {
...
+        Command::Init {
+            output_path,
+            dry_run,
+            quiet,
+            force,
+            no_input,
+            key_type,
+            key_file,
+            key_seed,
+            ..
```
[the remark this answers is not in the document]
QuinnWilton: Sure! That's probably the more correct choice anyway, since this will be the only consumer of that variant, so an unused field is likely a bug.

homestar-runtime/Cargo.toml
zeeshanlakhani: To @bgins below, my thought was that taplo, https://github.com/tamasfe/taplo, should handle this correctly, but I run taplo and haven't encountered these changes, so 🤷🏽.

QuinnWilton: Ah, my mistake, I dug into this a little bit more, and it looks like what's happening is that the `taplo` LSP is honoring the `tabSize` language server setting, which is being set by my editor to 4. I'll get rid of these formatting changes.

QuinnWilton:
I pushed up some small tweaks, including converting `InquireError` to a `miette` report, so that we're able to cleanup the error reporting on a SIGINT received during prompting. Here's how that looks now:

image

Exceptional errors still currently panic, which I think is possibly the right call for now, until we do the work of translating those into a `miette` error that directs users to file an issue in the repo or something.

bgins (approved):
Nice work on this! Looks great ✨

zeeshanlakhani:
@QuinnWilton could we add some integration/CLI (no runtime needed) tests to have tests around the functionality here?

homestar-runtime/Cargo.toml
```diff
@@ -80,6 +81,7 @@ http = "0.2"
 http-serde = "1.1"
 humantime = { workspace = true }
 indexmap = { workspace = true }
+inquire = "0.6.2"
```
zeeshanlakhani, line 84: minor: we let patches float up, so `0.6`.

homestar-runtime/Cargo.toml
```diff
@@ -54,6 +54,7 @@ crossbeam = "0.8"
 dagga = "0.2"
 dashmap = "5.5"
 derive-getters = "0.3"
+derive_builder = "0.13.0"
```
zeeshanlakhani, line 57: minor: we let patches float up, so `0.13`.

homestar-runtime/src/cli/init.rs
```diff
@@ -0,0 +1,317 @@
...
+    /// Generate the key from a seed
+    Seed {
+        /// The base64 encoded 32 byte seed to use for key generation
+        seed: Option<String>,
+    },
+}
+
+/// The type of key to generate
+#[derive(Debug, Clone, PartialEq, ValueEnum)]
+pub enum KeyTypeArg {
```
zeeshanlakhani, line 58: Throwing this out there: can we not reuse/extend KeyType from https://github.com/ipvm-wg/homestar/pull/582/files#diff-ee5055c2e074408bf2f7e2fbd583688042eaeec89be0fdeb760e7ca3cf2ef478R32?

homestar-runtime/src/cli/init.rs
```diff
@@ -0,0 +1,317 @@
...
+        match self {
+            PubkeyConfigOption::GenerateFromSeed => write!(f, "Generate from seed"),
+            PubkeyConfigOption::FromFile => write!(f, "From file"),
+        }
+    }
+}
+
+/// Handle the `init` command.
+pub fn handle_init_command(
+    output_mode: OutputMode,
```
zeeshanlakhani, line 93: minor: may be worth a struct argument vs multiple inputs?

homestar-runtime/src/cli/init.rs
```diff
@@ -0,0 +1,317 @@
+use std::{
+    fmt::Display,
+    fs::File,
+    io::{empty, stdout, Write},
+    path::PathBuf,
+    str::FromStr,
+};
+
```
zeeshanlakhani, line 8: minor: we let cargo fmt sort imports, so remove the space and it should auto-format accordingly.

homestar-runtime/src/settings.rs
```diff
@@ -352,13 +373,21 @@ impl Settings {
 }
 
 #[allow(dead_code)]
```
zeeshanlakhani, line 375: minor: allow/dead_code still needed w/ these updates?

homestar-runtime/src/cli/init.rs
```diff
@@ -0,0 +1,317 @@
...
+    no_input: bool,
+    _writer: &mut Box<dyn Write>,
+) -> Result<PubkeyConfig> {
+    let config = match key_arg {
+        None => {
+            if no_input {
+                bail!("Aborting... cannot prompt for key in non-interactive mode. Pass`--key-file <KEY_FILE>` or `--key-seed [<KEY_SEED>]` to configure the key.");
+            }
+
+            let options = vec![
```
zeeshanlakhani, line 252: can this be made a const array? Or does Select for the vec anyway?

homestar-runtime/src/cli/init.rs
zeeshanlakhani: Not doing a deep-dive yet on pubkey_config again, but are there are more shared elements with a little bit of cleanup? Throwing that out there just as a check.

homestar-runtime/src/cli/init.rs
```diff
@@ -0,0 +1,317 @@
...
+                .read(true)
+                .write(true)
+                .create_new(true)
+                .open(&path);
+
+            // This seemingly convoluted match is to avoid the risk of a
+            // TOCTOU race condition, where another process creates the file
+            // in between this one checking for its existence and opening it.
+            //
+            // TODO: there should probably be a flag for non-interactive use
```
zeeshanlakhani, line 178: we do have a --no-input flag for non-interactive use?

zeeshanlakhani (commented):
Added some comments thus far.

- [ ] Q: Shouldn't we also have the option to generate a PEM/DER file for users (not just provide your own) as another option? Essentially, replace the OpenSSL genkey we've/users have been doing manually? I thought this was one goal of the work.

homestar-runtime/src/cli/init.rs
QuinnWilton: I was torn on reusing `KeyType`, because that would give up control over which options are displayed in the initialization wizard, when I can imagine situations where we may want to support legacy key types (like RSA), but without necessarily exposing them through the first-time setup.

Happy to reuse it if you don't think that's a concern though!

zeeshanlakhani: I figure that since pubkey config will always be tied to what a user can provide (i.e. settings), it'd be in sync with whichever key types we'll need to init with?

homestar-runtime/src/cli/init.rs
QuinnWilton: Good catch, thanks!

homestar-runtime/src/cli/init.rs
QuinnWilton: I don't believe so, unfortunately. It moves the selected option to return it by ownership, so we'd also need to implement `Copy` over the `KeyType`, but even then, I'm only seeing a constructor over a `Vec`

homestar-runtime/src/settings.rs
QuinnWilton: Nope! Thanks :)

QuinnWilton:
> Added some comments thus far.
> 
> * [ ]  Q: Shouldn't we also have the option to generate a PEM/DER file for users (not just provide your own) as another option? Essentially, replace the OpenSSL genkey we've/users have been doing manually? I thought this was one goal of the work?
> 
> ❤️ the experience btw.

I don't see any APIs exposed by the `libp2p-identity` crate for encoding keys as PEM/DER files. Are you aware of anything that I may have missed? Or were you imagining that we'd pull in another crypto crate / shell out to `OpenSSL` for that?

Either way, I just pushed up fixes for all of your other comments.

QuinnWilton:
@zeeshanlakhani I just pushed up PEM generation for ed25519, improved PEM parsing for ed25519, and the tests you requested! Headed to dinner now, but merge conflicts are fixed and I'll check CI in the morning 😊

homestar-runtime/src/cli/init.rs
```diff
@@ -0,0 +1,362 @@
...
+                        .with_default_value_formatter(&|_| "random".to_string())
+                        .with_error_message("Please type a base64 encoding of 32 bytes")
+                        .with_help_message("Base64 encoded 32 bytes")
+                        .prompt()
+                        .map_err(|e| miette!(e))?;
+
+                    PubkeyConfig::GenerateFromSeed(RNGSeed::new(key_type, seed.0))
+                }
+                PubkeyConfigOption::FromFile => {
+                    // HACK: We need to manually instantiate the struct with a custom formatter,
```
zeeshanlakhani, line 280: not sure we need to call this a hack vs constraints on the type/impls?

zeeshanlakhani: I guess my other Q is if we had to use PathBuf, but I'm guessing yes.

zeeshanlakhani:
@QuinnWilton Code is looking pretty good, but there was a new setting added in `main` that's missing here after rebase. Once that's updated, I'll do a run through and check the DX experience.

homestar-runtime/src/cli/init.rs
zeeshanlakhani: @QuinnWilton yeah, seeing your responses, I think we can rip this out easily and go with String.

QuinnWilton: Cool, I'll do that, thanks!

bgins:
Hey @QuinnWilton! Trying out the new PEM key file features. They work great, and I love them. ❤️ 

I have a couple of minor DX requests:

- Could we update the help text to indicate that the `--key-file` flag will generate a key if an existing key does not exist at the path?
- When generating a key file with the `--key-file` flag and selecting the `Secp256k1` option, an unsupported error is thrown (as expected) but an empty key file is created. Could we clean up and remove the key file in this error case?

zeeshanlakhani:
@QuinnWilton last minor thing:

- Could we generate the pem in the config directory by default if no file is given to key-file and override the default path if run again?

homestar-runtime/src/cli.rs
```diff
@@ -34,12 +37,84 @@ Usage: {usage}
...
+        long = "key-type",
+        value_name = "KEY_TYPE",
+        help = "The type of key to use for libp2p [optional]"
+    )]
+    pub key_type: Option<KeyType>,
+    /// The file to load the key from
+    #[arg(
+        long = "key-file",
+        value_name = "KEY_FILE",
+        help = "The path to the key file. A key will be generated if the file does not exist [optional]",
```
zeeshanlakhani, line 104: minor: add a note that a key will be generated to <default path> if no file is provided.

homestar-runtime/src/cli/init.rs
```diff
@@ -0,0 +1,366 @@
...
+                        .read(true)
+                        .write(true)
+                        .create_new(false)
+                        .open(&path)
+                        .expect("to open settings file")
+                }
+                err => err.expect("to open settings file"),
+            };
+
+            writeln!(writer, "Writing settings to {:?}", path).expect("to write");
```
zeeshanlakhani, line 214: b/c we could also be writing out the PEM (whether defaulted or given), can we have a note to the console saying that? The case I realized:

didn't overwrite settings, but pem was written in the previous step without any ack.

zeeshanlakhani (approved):
@QuinnWilton left a few last small changes/details. Great work.

homestar-runtime/src/cli/init.rs
```diff
@@ -0,0 +1,372 @@
...
+                .expect("to write to key file");
+        }
+        // file did exist, do nothing and use existing key
+        Err(err) if err.kind() == std::io::ErrorKind::AlreadyExists => {}
+        err => {
+            err.expect("to open key file");
+        }
+    };
+
+    writeln!(writer, "Writing key file to {:?}", path).expect("to write");
```
bgins, line 369: Hey! Testing out the latest changes. It looks like this message will be logged when using an existing key, passed in through the `--key-file` option. Would it be possible to log something like "Using existing key..." here in this case?

QuinnWilton: Ah whoops, you're totally right. That's what I get for cranking that out while making lunch 😅 Thanks!

QuinnWilton: Fixed in latest push: 
image

bgins: Nice! Looks great ❤️

PR #594, opened 2024-02-29
chore: Add release-plz schema action run

bgins:
# Description

This PR makes the following changes:

- [x] Update schemas action to run on `release-plz` branches on manifest changes only
- [x] Update schemas action to run on `main` for both manifest and Rust changes

`release-plz` updates versions that we reference in schemas, which put the versions in the schemas out of date. This PR updates the action to handle version changes.

## Type of change

### DIFF
--- a/.github/workflows/schemas.yml
+++ b/.github/workflows/schemas.yml
@@ -2,7 +2,8 @@ name: 📄 Schemas
 
 on:
   push:
-    branches: [main, "release-plz-*"]
+    branches: [main, "bgins/run-schemas-*"]
+    # branches: [main, "release-plz-*"]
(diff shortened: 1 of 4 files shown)


.github/workflows/schemas.yml
```diff
@@ -58,12 +59,17 @@ jobs:
         run: cargo run -p homestar-schemas
 
       - name: Check for modified schemas
-        id: git-check
+        id: git-check-schemas
```
zeeshanlakhani, line 62: let's run the schema one only on `push` to `main`.

zeeshanlakhani (approved):

Issue #457, opened 2023-11-29
Add workflow and runner instrumentation and traces


bgins:
# Summary

Let's add workflow instrumentation and tracing to track more information about workflow runs.

# Detail

We should add instrumentation and tracing to determine:

- [ ] Total runs for a workflow
- [ ] Total ok/errors within a workflow run
- [ ] Total tasks replayed within a workflow
- [ ] Time spent running a workflow
- [ ] Time spent running tasks within a workflow

PR #603, opened 2024-03-07
refactor: Add workflow spans and every cli logging

bgins:
# Description

- [x] Add initial workflow and task execution spans
- [x] Add workflow initialize, start, and end logs
- [x] Add computed and replayed receipt logs
- [x] Add custom logging format and filter for EveryCLI
- [x] Minor re-wording of existing logs

## Link to issue

Implements spans needed in #457

## Type of change

## Test plan (required)

Run Homestar with `EVERY_CLI` set to `true` to see the simplified logs:

```
EVERY_CLI=true cargo run -- start
```

Check that `RUST_LOG` works when `EVERY_CLI` is `false` or not set:

 ```
EVERY_CLI=false RUST_LOG=info cargo run -- start
```

## Screenshots

Running `test-workflow-add-one.json` with `EVERY_CLI` logs:

CleanShot 2024-03-12 at 22 10 28@2x

Same workflow, on replay:

CleanShot 2024-03-12 at 22 12 12@2x

A different workflow with WASI logging at `trace` level:

CleanShot 2024-03-12 at 22 17 42@2x

Error case where an arg with an unexpected type is passed to a workflow task:

CleanShot 2024-03-12 at 22 14 17@2x

### DIFF
--- a/homestar-runtime/src/logger.rs
+++ b/homestar-runtime/src/logger.rs
@@ -7,6 +7,7 @@ use tracing_subscriber::{layer::SubscriberExt as _, prelude::*, EnvFilter};
 
 const LOG_FILE: &str = "homestar.log";
 const DIRECTIVE_EXPECT: &str = "Invalid tracing directive";
+const EVERY_CLI: &str = "EVERY_CLI";
 
 /// Logger interface.
 #[derive(Debug)]
@@ -44,7 +45,7 @@ fn init(
     #[allow(unused_variables)] settings: &settings::Monitoring,
 ) -> WorkerGuard {
     // RUST_LOG ignored when EVERY_CLI is true
-    let every_cli = std::env::var("EVERY_CLI").is_ok_and(|val| val == "true");
+    let every_cli: bool = std::env::var(EVERY_CLI).is_ok_and(|val| val == "true");
 
     // TODO: Add support for customizing logger(s) / specialzed formatters.
     let format_layer = if every_cli {
--- a/homestar-runtime/src/logger.rs
+++ b/homestar-runtime/src/logger.rs
@@ -7,6 +7,7 @@ use tracing_subscriber::{layer::SubscriberExt as _, prelude::*, EnvFilter};
 
 const LOG_FILE: &str = "homestar.log";
 const DIRECTIVE_EXPECT: &str = "Invalid tracing directive";
+// Sets simplified logging filter and format for Every CLI
 const EVERY_CLI: &str = "EVERY_CLI";
 
 /// Logger interface.
--- a/homestar-runtime/src/logger.rs
+++ b/homestar-runtime/src/logger.rs
@@ -73,6 +73,11 @@ fn init(
     let filter = if every_cli {
         EnvFilter::new("off")
             .add_directive(
+                "homestar_runtime::runner[run_worker]=info"
+                    .parse()
+                    .expect(DIRECTIVE_EXPECT),
+            )
+            .add_directive(
                 "homestar_runtime::worker[run]=info"
                     .parse()
                     .expect(DIRECTIVE_EXPECT),
--- a/homestar-runtime/src/logger.rs
+++ b/homestar-runtime/src/logger.rs
@@ -81,6 +81,11 @@ fn init(
                     .parse()
                     .expect(DIRECTIVE_EXPECT),
             )
+            .add_directive(
+                "homestar_wasm[wasi_log]=trace"
+                    .parse()
+                    .expect(DIRECTIVE_EXPECT),
+            )
     } else {
         EnvFilter::try_from_default_env().unwrap_or_else(|_| {
             EnvFilter::new("info")
--- a/homestar-runtime/src/logger.rs
+++ b/homestar-runtime/src/logger.rs
@@ -77,7 +77,7 @@ fn init(
                     .expect(DIRECTIVE_EXPECT),
             )
             .add_directive(
-                "homestar_runtime::worker[spawn_tasks]=info"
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: homestar-runtime/src/worker.rs [hunk 1]
```diff
@@ -209,8 +219,26 @@ where
                     )?;
                 }
 
+                info!(
```
zeeshanlakhani, line 222: -> move this to where we complete status at the end of batches.

homestar-runtime/src/worker.rs
```diff
@@ -209,8 +219,26 @@ where
...
+                    workflow_cid,
+                    "starting workflow"
+                );
+
                 // Run the queue of tasks.
-                self.run_queue(ctx.scheduler, running_tasks).await
+                let result = self.run_queue(ctx.scheduler, running_tasks).await;
+
+                if let Ok(()) = result {
+                    info!(
```
zeeshanlakhani, line 233: -> move this to where we complete the status at the end of batches.

Have the log captured by the span for replaying receipts, which will happen whether or not the workflow is complete or not.

homestar-runtime/src/worker.rs
```diff
@@ -352,7 +383,11 @@ where
                                         })
                                 }
                             }
-                        });
+                        }
+                        .instrument({
+                            info_span!("spawn_tasks").or_current()
```
zeeshanlakhani, line 388: renaming this span to have more helpful information than `spawn tasks`.

Spawning workflow tasks or something like this.

zeeshanlakhani: Maybe add the workflow cid to this span info.

homestar-runtime/src/worker.rs
```diff
@@ -321,17 +350,19 @@ where
...
                                 .boxed()
                         });
 
                         let handle = task_set.spawn(async move {
-                            match resolved.await {
+                             match resolved.await {
                                 Ok(inst_result) => {
-                                    match wasm_ctx.run(wasm, &fun, inst_result).await {
+                                    match wasm_ctx.run(wasm, &fun, inst_result).instrument({
+                                        info_span!("wasm_run").or_current()
```
zeeshanlakhani, line 364: check: should this be a debug_span?

homestar-runtime/src/logger.rs
```diff
@@ -43,31 +43,60 @@ fn init(
...
+            .layer()
+            .with_writer(writer)
+    };
 
-    let filter = EnvFilter::try_from_default_env().unwrap_or_else(|_| {
-        EnvFilter::new("info")
-            .add_directive("homestar_wasm=info".parse().expect(DIRECTIVE_EXPECT))
-            .add_directive("libp2p=info".parse().expect(DIRECTIVE_EXPECT))
+    let filter = if every_cli {
+        EnvFilter::new("off")
```
zeeshanlakhani, line 73: My only possible qualm here is we won't get general errors (we can skip all deps ones) that could be helpful at the highest level (remembering that `error`) is the highest level. Nonetheless, I won't be a stickler here, but maybe worth chatting with @avivash?

homestar-runtime/src/worker.rs
bgins: Done 👌

homestar-runtime/src/logger.rs
```diff
@@ -43,31 +43,65 @@ fn init(
     guard: WorkerGuard,
     #[allow(unused_variables)] settings: &settings::Monitoring,
 ) -> WorkerGuard {
+    // RUST_LOG ignored when EVERY_CLI is true
+    let every_cli = std::env::var("EVERY_CLI").is_ok_and(|val| val == "true");
```
zeeshanlakhani, line 47: minor, can we const this at the top.

zeeshanlakhani (approved):
just a minor ask and that debug question remains, but it's not necessary. 👍🏽

homestar-runtime/src/worker.rs
bgins: Checked this, works fine 👌

homestar-runtime/src/logger.rs
bgins: Yep done 👍

homestar-runtime/src/logger.rs
```diff
@@ -7,6 +7,7 @@ use tracing_subscriber::{layer::SubscriberExt as _, prelude::*, EnvFilter};
 
 const LOG_FILE: &str = "homestar.log";
 const DIRECTIVE_EXPECT: &str = "Invalid tracing directive";
+const EVERY_CLI: &str = "EVERY_CLI";
```
zeeshanlakhani, line 10: minor: I'd probably write a small comment at the top on any env vars and why they're generally here.

homestar-wasm/src/wasmtime/world.rs
```diff
@@ -145,6 +146,7 @@ impl<T> Env<T> {
     /// Types must conform to [Wit] IDL types when Wasm was compiled/generated.
     ///
     /// [Wit]: <https://github.com/WebAssembly/component-model/blob/main/design/mvp/WIT.md>
+    #[instrument(name = "execute", skip_all)]
```
zeeshanlakhani, line 149: minor: does this need the name since it matches the fn? I thought that was automatic, but maybe not?

zeeshanlakhani (approved):
A couple super minor comments, otherwise 👍🏽

homestar-runtime/src/logger.rs
bgins: Yep, adding a comment 👍

